PR #8, opened 2017-08-15
Create redesign-proposal.md

munnerz:


docs/redesign-proposal.md
```diff
@@ -0,0 +1,93 @@
...
+cluster). This may include things like notifications that a scale down event
+has been requested, or even used to elect a 'leader' between the Pilots in the
+event one is required for the proper functioning of that application.
+
+Each Pilot should also be able to expose a metrics endpoint to be scraped by
+systems like Prometheus. This could be implemented via some of the existing
+exporters for common applications, or directly within the Pilot itself if
+required.
+
+## navigator-controller-manager
```
simonswine: I think we should we mention something around helm/tiller integration points here

simonswine (approved):
Good work james, nothing to object. I think it would be good to keep the idea of helm as a generic concept to template resources as port of the controller manager

docs/redesign-proposal.md
```diff
@@ -0,0 +1,93 @@
...
+
+If a user were to modify the size of a node pool within an
+ElasticsearchCluster, the controller-manager would consequently notify the
+relevant Pilots of this event. It could then wait until a successful data
+migration has occured before actually scaling down the appropriate
+StatefulSet/Deployment. The precise semantics of *how* the controller-manager
+should communicate this to the Pilots is undefined, but however it is done it
+should be via the API server (either by updating the `status` field of the
+ElasticsearchCluster resource, or by creating an additional `ScaleEvent`
+resource to be fulfilled by the Pilots).
```
wallrj: ❓ Maybe add a little more background here. Why is this necessary? Why not just have the navigator-controller update the `StatefulSet.replicas` value and let Kubernetes do the rest?

docs/redesign-proposal.md
```diff
@@ -0,0 +1,93 @@
...
+manage complex systems on top of cluster schedulers, initially targetting
+Kubernetes and OpenShift.
+
+## Architecture
+
+All communication between components of the Navigator system should communicate
+through the `navigator-apiserver`. This allows for a consistent audit trail,
+recorded in a central place. Monitoring services such as `kube-state-metrics`
+can then be used to track changes to the contents of the API server to be
+consumed by something like Prometheus.
```
wallrj: Having read further, it seems that one key reason for `navigator-api-server` is that OpenShift (and other OFs) don't support custom resource definitions. In that case, maybe state it here.

docs/redesign-proposal.md
```diff
@@ -0,0 +1,93 @@
...
+StatefulSet/Deployment. The precise semantics of *how* the controller-manager
+should communicate this to the Pilots is undefined, but however it is done it
+should be via the API server (either by updating the `status` field of the
+ElasticsearchCluster resource, or by creating an additional `ScaleEvent`
+resource to be fulfilled by the Pilots).
+
+Only one instance of the controller-manager should run at a given time, and
+they should leader-elect amongst themselves using either the
+`navigator-apiserver` or some other leader election primitive provided by the
+target orchestrator.
```
wallrj: ❓ Why only one instance? Isn't there supposed to be a controller instance for each database type?

wallrj: ❓ And isn't leader election for controllers built into Kubernetes somehow? How does the Kubernetes StatefulSet controller ensure that only the controller on a single master responds to StatefulSet changes?

docs/redesign-proposal.md
```diff
@@ -0,0 +1,93 @@
...
+
+## navigator-apiserver
+
+The apiserver component acts as the primary API surface and data storage for
+all resources types that we define as part of our API. It's design should be
+heavily modelled upon that of `kube-apiserver`, in that it should not perform
+much actual control logic itself.
```
wallrj: ❓ Will it perform *any* control logic? If so, what?

docs/redesign-proposal.md
```diff
@@ -0,0 +1,93 @@
...
+allowing us to support `v1` configuration for types even with the newest
+versions of the `navigator-controller-manager` by losslessly converting between
+the old and new version types. More information about this can be found in
+my talk on API internals, as well as on the upcoming blog post.
+
+Like the `kube-apiserver`, multiple instances of it may run at one time to
+provide scalability on both reads and writes. The underlying data store may be
+`etcd`, Kubernetes CRDs or OpenShift TPRs. It is undecided which will be
+implemented first, however good support for `etcd` is provided in the
+`k8s.io/apiserver` library.
```
wallrj: If not custom resources or third party resource, will users  be able to use `kubectl` to interact with it?

wallrj (commented):
Thanks @munnerz 

It mostly makes sense, but personally I'd need a bit more context to fully understand how you've reached this design.

If you have time I'd be interested to know:

1. What lessons were learned from previous ElasticSearch Statefulset experiments?
2. What lessons were learned from the Couchbase-sidecar experiment.
3. What were the shortcomings of the original navigator architecture.
4. How is this architecture compare to ElasticSearch  and Prometheus operators from CoreOS.
5. How is OpenShift approaching this problem? They have built in support for ElasticSearch and other databases right? How is navigator better than those?
6. Apart from Openshift and Kubernetes, which other orchestration frameworks do you imagine we could support?
7. I've heard you talk about navigator-controller interacting with helm / tiller ....deploying helm charts rather than creating StatefulSet resources directly. Where as my understanding is that developers / end users want to use helm to deploy their applications to Kubernetes. Will I be able to write a helm chart that creates an ElasticSearchDeployment via the navigator API and also creates a Kibana / LogStash Deployment via the Kubernetes API?

That's all for now. 
Thanks again for writing all this down, it's been most helpful for me to understand the big picture.

docs/redesign-proposal.md
munnerz: @wallrj 

(1) Regarding the audit trail, this is undecided at present. CustomResourceDefinitions and ThirdPartyResources should be thought of as more of a 'simple database' than anything else. Whilst `kubectl` allows you to interact with them, it does no more other than CRUD operations. If the user is using CRD/TPRs as a storage backend for the navigator-apiserver, then yes there will be CRUD events logged in whatever audit framework the `kube-apiserver` has, however a user may *not* be using CRD/TPR as a storage backend, and regardless Navigator itself should manage this so that it can provide the same functionality between different environments (say, in an environment that doesn't use containers, ie. plain GCE or AWS).

(2) Just like `kube-apiserver`, there is one, although it can scale horizontally (again, like the `kube-apiserver`). The reason they can scale horizontally is that the backing storage (ie. CRDs or TPRs or etcd) is responsible for ensuring consistency.

(3) Access control is initially out of scope. If using the kube-aggregator (in the Kubernetes case) to expose the `navigator-apiserver`, 'delegated authentication' can be used (ie. relying on *another* API server to handle auth). More details/an example is available in the `service-catalog` repo: https://github.com/kubernetes-incubator/service-catalog/blob/master/docs/auth.md

(4) Not just this lack of support for CRDs, but we may at some point be working in an environment completely unlike K8s/OS (ie. targeting a cloud directly perhaps). But yep, worth mentioning.

docs/redesign-proposal.md
```diff
@@ -0,0 +1,93 @@
...
+
+All communication between components of the Navigator system should communicate
+through the `navigator-apiserver`. This allows for a consistent audit trail,
+recorded in a central place. Monitoring services such as `kube-state-metrics`
+can then be used to track changes to the contents of the API server to be
+consumed by something like Prometheus.
+
+## Pilots
+
+Databases run by Navigator are wrapped by helper processes called 'Pilots'.
```
munnerz: Ack - I'll update to mentioning forking a new subprocess.

docs/redesign-proposal.md
munnerz: (1) No, there is one controller manager for all at the moment, like how there is one controller manager for all k8s resource types. In future we could consider breaking this out into multiple standalone binaries, if there is a pressing reason to.

(2) Yes it is, and we *could* consider using it, however that may also put a dependence on running in an environment that already provides leader election primitives. By relying *only* on our own apiserver component, we can be sure that this will run outside of K8s/OS

docs/redesign-proposal.md
munnerz: No it won't - the closest thing to 'control logic' is defaulting fields that aren't set. Aside from that, the apiserver should be a 'dumb' file store.

docs/redesign-proposal.md
munnerz: Yes - by talking to the `navigator-apiserver` directly or, in the case of Kubernetes (and potentially OpenShift in future), by talking to the `kube-aggregator`, which will proxy the request through to `navigator-apiserver`.

munnerz: This is as a consequence of us following the K8s API conventions when building our API, hence us basing our own apiserver off the `github.com/kubernetes/apiserver` library.

docs/redesign-proposal.md
munnerz: So in the event of a scale down event, it's important that the Pilot is aware it's *going* to happen so it can take some action to mitigate it. *Sometimes*, this may be to just do nothing. But in some cases we might want to evict all data/shards off of that replica. In order to not be subject to the 'killing policy' of the orchestrator (ie. kubernetes `terminationGracePeriodSeconds`), it's useful to have 'fore-warning' of these events so that a data migration can begin *before* the countdown for `terminationGracePeriodSeconds` begins.

docs/redesign-proposal.md
```diff
@@ -0,0 +1,93 @@
...
+to trap signals sent to the container and ensure the underlying application
+exits cleanly, and if it doesn't, report this diagnostic information back to
+the `navigator-apiserver`.
+
+Pilots should watch the `navigator-apiserver` for changes to resources it
+requires, and use the apiserver to communicate with other instances of Pilot
+within the same application cluster (eg. within the same Elasticsearch
+cluster). This may include things like notifications that a scale down event
+has been requested, or even used to elect a 'leader' between the Pilots in the
+event one is required for the proper functioning of that application.
```
munnerz: * kubectl create -f elasticsearch1.yaml`
* `navigator-apiserver` stores resource
* `navigator-controller-manager` watches and responds to new ElasticsearchCluster resources
* `navigator-controller-manager` creates StatefulSet resources (+ anything else needed to fulfil the ElasticsearchCluster)
* ` kubernetes` creates pods & PVs for StatefulSet
* `Pod1` starts up
* `ElasticSearchPilot` connects to navigator-api and attempts to claim the leader lock

Now it's worth noting here that the lock that the Pilot acquires **IS NOT** the same as the Elasticsearch leadership election. It may be the case that Elasticsearch and Pilot leaders are in different pods. The Pilots will elect a leader *if they need to*. So in the Elasticsearch case, a leader is chosen to be the one instance of Pilot that sends cluster control events (such as draining a node of shards) eg. in response to a scale down. That leaders may in fact be different to whatever leader Elasticsearch has elected.

Again though, this is TBD and not solidified yet. & is an implementation detail of the Pilot itself (ie. this should not be Navigators concern directly). As far as Navigator is concerned, there's no such thing as a leader really. It provides a primitive that can be used by applications to *determine* a leader, but neither the `navigator-apiserver` or `navigator-controller-manager` should care whether or not a leader exists. Instead, each Pilot decides whether to take some action (eg. draining shards from a node) depending whether or not it is a leader.

PR #16, opened 2017-08-31
Switch to new k8s.io/code-generator repo

munnerz:
This PR switches us to use the new code-generator repo. In the process, we have updated the version of the code generator used.


munnerz:
Rebased after other PRs have merged to master. Ready to look over now @wallrj, although may be a bit hard!

munnerz:
Now that we have e2e tests however, it should be easier to verify it works!

Gopkg.toml
```diff
@@ -20,7 +20,27 @@
...
 #  version = "2.4.0"
 
+required = [
+  "k8s.io/code-generator/cmd/client-gen",
+  "k8s.io/code-generator/cmd/conversion-gen",
+  "k8s.io/code-generator/cmd/deepcopy-gen",
+  "k8s.io/code-generator/cmd/defaulter-gen",
+  "k8s.io/code-generator/cmd/informer-gen",
+  "k8s.io/code-generator/cmd/lister-gen",
+]
```
wallrj: ❓ Why list all the commands rather than the parent package?
```
required  = [
    "k8s.io/code-generator",
]
```

hack/boilerplate.go.txt
```diff
@@ -0,0 +1,15 @@
+/*
+Copyright YEAR The Kubernetes Authors.
```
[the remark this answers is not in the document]
munnerz: Probably, yes 😄

Gopkg.toml
munnerz: We don't require the entire parent package, we require just the dependencies of the actual binaries themselves. If we depend on the parent package, we'll pull in a lot more than is necessary.

Shortly I'm going to follow up with a PR after running `dep prune`, that removes a *lot* of unused packages. If I required the parent, it would not remove some of those despite them not being used 😄

munnerz:
@wallrj there are a lot of changes because as a result of this, I've had to updated the apimachinery, client-go dependencies as well as adding a dependency on k8s.io/api. Consequently, *every* API type has moved, which will cause a *tonne* of churn on line numbers.

PR #25, opened 2017-09-05
Implement custom API server with kube-aggregator

munnerz:
This PR is an initial pass at creating a custom API server that registers with kube-aggregator.

It currently works, and I've managed to run it all with the Navigator controller.

This PR is WIP as it still needs supporting CI infrastructure (testing, building, pushing images) before it can be merged.

Due to some version conflicts with a few dependencies (specifically `github.com/coreos/etcd` and `google.golang.org/grpc`) and the need to vendor the latest `k8s.io/apiserver`, quite a few dependencies have been touched in this PR.


munnerz:
😲

screen shot 2017-09-06 at 00 15 48

munnerz:
Hey - thanks for the comments!

Yep it is a large PR thanks to vendor changes - I tend to just review it commit-by-commit instead of as a whole, as it also gives me an idea of the thinking behind each step.

> navigator/pkg/apiserver has the Kubernetes copyright boilerplate rather than Jetstack boilerplate.
> We should consider automatically checking and updating the boilerplate for our opensource projects, like Kubernetes does.

The copyright boilerplate was taken from the `k8s.io/sample-apiserver` repo, as this is what I based a fair bit of the code on. Perhaps it *is* okay to switch it to Jetstack though.

>I was imagining that we'd have a separate API server for each database type.
> That way we'd be able to bundle all the Couchbase navigator components in a single repo.

At the moment we will have one API server for all the database types. I don't see a reason to split it out, and I think it will create extra work for us. In future if there is a strong reason to, I'm happy to reconsider this though. The API registration API does allow us to do this quite easily.

Regarding storing all Couchbase related stuff in a single repo - everything related to actually interacting with Couchbase *does* stay in a single repo, pilot-couchbase. This PR hasn't changed much in terms of what code is stored in the navigator repo. It just sets up a RESTful server to serve our API types up and perform conversions/defaults etc. between API versions within our group.

> It's a shame to disable the e2e tests.
> How have you tested this? How do you interact with the custom API server if not using kubectl?

I think you've misunderstood how the custom API server works - we *do* use kubectl to interact with it. The API server gets registered with kube-aggregator which then proxies requests to our API server for our particular API group/version. You can see the resource that triggers the registration here: https://github.com/jetstack-experimental/navigator/blob/impl-apiserver/contrib/charts/navigator/templates/apiservice.yaml. There's not really (m)any docs on how to do this at the moment, but there's a skeleton here: https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/

> I see the custom API server listens on port 443
> But I can't see how you discover the IP address of the API server?
> I imagined that the main Kubernetes API would reverse proxy requests to navigator API URL > segments, to the custom API server.
> Is that how it works?
> And if so, why can't we use kubectl to interact with it in the e2e tests?

Yes that is how it works - and we can use kubectl to interact with it.

The reason for disabling the e2e tests was not being able to change image tags, and the changes in images being made making things more complex. I've started work to re-enable them by updating the helm chart to include the apiserver component.

> Is there a way to generate REST API docs for the custom API server?
> Are those included in this PR?

Yes there is, with openapi-gen. That will be a separate PR however.

> How do the generated Navigator informers and listers (or the underlying client) connect to the custom API?

Nothing changes here - we simply connect to the Kubernetes API server and it'll proxy requests appropriately.


----

> will having a separate API server make Navigator more portable to other orchestration frameworks?

So whilst in this comment I've talked about kube-aggregator, as you've noticed, our apiserver ultimately does just listen on port 443 and server requests. This means that we can always *bypass* the aggregator and talk directly to our custom API server. In environments that don't support kubernetes API aggregation, we can have users talk directly to our API server instead of via kube-aggregator (by reconfiguring their kubectl to talk to our server). As `navigator-apiserver` is a fully compliant Kubernetes API server, kubectl will be able to discover the API types we support and allow users to query those.

> Do Docker swarm or Mesosphere have a concept of a Statefulset? And if not, will we aim to implement all that in Navigator?

We will have to assess each one as they come. This PR does not aim to resolve multi-orchestrator support in the controller component, it simply looks to introduce the apiserver component.

> And by making Navigator OF agnostic, will we end up making it too complicated for Kubernetes users to contribute plugins for other databases?

Good point - it'd be great to get this as it's own issue, as we should try to come up with a way to 'resolve' this...

PR #39, opened 2017-10-02
Add Pilot resource type

munnerz:


munnerz:
Merging this as it's basically master now given it blocks everything.

This now stands up functional clusters. I need to improve the status fields reported back on the Pilot resource, and add a leader elected sync loop to the genericpilot in order to handle scale down events.

PR #45, opened 2017-10-17
Various fixes for running e2e tests locally

wallrj:
* Simplify the navigator_ready function to only check for api-versions
* Tidy up the e2e prepare script (remove duplication and unused variables)
* Remove operations from travis.yaml which are duplicated in the e2e prepare script.

Regarding the original problems that I reported in https://gitlab.jetstack.net/marshal/navigator/issues/27

> On my local Minikube, the navigator APIs take a long time to be registered...up to 5 minutes.

This seems to be caused by kube-dns restarting due to RBAC until its service account has been created.
If I wait for kube-dns to settle down the API appears quickly and tests run reliably.

And this problem is exacerbated by Disk IO failures inside the minikube virtualbox VM.

I switched to using the kvm driver and now the tests pass reliably.

> Additionally the check for Deployment.status.readyReplicas no longer works.

Maybe this was caused by my using kubectl 1.6?
I upgraded to `kubectl v1.8` and the readyReplicas attribute is now reported, but nevertheless, I've modified the code to simply check for the availability of the navigator API.

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -47,8 +47,6 @@ function fail_test() {
     kubectl describe svc
     kubectl get apiservice -o yaml
     kubectl describe apiservice
-    kubectl logs -c apiserver -l app=navigator,component=apiserver
-    kubectl logs -c controller -l app=navigator,component=controller
     exit 1
 }
 
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -32,9 +32,14 @@ function navigator_ready() {
     return 1
 }
 
-echo "Waiting up to 10 minutes for Navigator to be ready..."
-if ! retry TIMEOUT=600 navigator_ready; then
-    kubectl api-versions
+echo "Waiting for Navigator to be ready..."
+if ! retry navigator_ready; then
+    (
+        kubectl api-versions
+        kubectl get pods --all-namespaces
+        kubectl describe deploy
+        kubectl describe pod
+    ) > debug.log
     echo "ERROR: Timeout waiting for Navigator API"
     exit 1
 fi
@@ -50,7 +55,7 @@ function fail_test() {
 
 function test_elasticsearchcluster() {
     echo "Testing ElasticsearchCluster"
-    if ! retry kubectl get esc; then
+    if ! kubectl get esc; then
         fail_test "Failed to use shortname to get ElasticsearchClusters"
     fi
     # Create and delete an ElasticSearchCluster
@@ -73,5 +78,18 @@ function test_elasticsearchcluster() {
 }
 
 test_elasticsearchcluster
+
+if [[ "${FAILURE_COUNT}" -gt 0 ]]; then
+    (
+        kubectl get po -o yaml
+        kubectl describe po
+        kubectl get svc -o yaml
+        kubectl describe svc
+        kubectl get apiservice -o yaml
+        kubectl describe apiservice
+        kubectl logs -c apiserver -l app=navigator,component=apiserver
+        kubectl logs -c controller -l app=navigator,component=controller
+    ) > debug.log
+fi
 
 exit $FAILURE_COUNT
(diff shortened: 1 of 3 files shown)


hack/e2e.sh
```diff
@@ -19,57 +19,26 @@ source "${SCRIPT_DIR}/libe2e.sh"
...
     return 0
 }
 
-if ! retry navigator_ready; then
-        kubectl get pods --all-namespaces
-        kubectl describe deploy
-        kubectl describe pod
-        exit 1
+echo "Waiting up to 10 minutes for Navigator to be ready..."
+if ! retry TIMEOUT=600 navigator_ready; then
```
munnerz: I'd class the API taking 10 minutes to register as a bug and not something we should accept - can we remove this change from this PR?

Happy to lend a hand debugging what's going on your end to cause those issues though. Let's discuss and work through it on Slack.

hack/e2e.sh
```diff
@@ -19,57 +19,26 @@ source "${SCRIPT_DIR}/libe2e.sh"
...
-    fi
-    local replica_count_apiserver=$(
-        kubectl get deployment ${RELEASE_NAME}-navigator-apiserver \
-                --output 'jsonpath={.status.readyReplicas}' || true)
-    if [[ "${replica_count_apiserver}" -eq 0 ]]; then
-        return 1
+    if kubectl api-versions | grep 'navigator.jetstack.io'; then
+        return 0
     fi
     return 0
```
munnerz: This should be 'return 1' if the above is 'return 0. This function cannot fail at the moment.

hack/e2e.sh
```diff
@@ -19,57 +19,26 @@ source "${SCRIPT_DIR}/libe2e.sh"
...
-if ! retry navigator_ready; then
-        kubectl get pods --all-namespaces
-        kubectl describe deploy
-        kubectl describe pod
-        exit 1
+echo "Waiting up to 10 minutes for Navigator to be ready..."
+if ! retry TIMEOUT=600 navigator_ready; then
+    kubectl api-versions
+    echo "ERROR: Timeout waiting for Navigator API"
+    exit 1
```
munnerz: I include the `describe` lines in here to make it possible to debug. Any reason why you've removed?

hack/e2e.sh
wallrj: It seems to be related to RBAC.
If I start minikube without rbac enabled and disable rbac in the navigator chart, the Navigator API becomes ready within 3 minutes.

hack/e2e.sh
wallrj: The tests kept failing and I couldn't see the wood for the trees.
Can we put this in a debug.log file instead?

wallrj:
I've addressed your comments @munnerz  and updated the description now that I (think) I've got to the bottom of the problems.

munnerz:
> I upgraded to kubectl v1.8 and the readyReplicas attribute is now reported, but nevertheless, I've modified the code to simply check for the availability of the navigator API.

Can we switch this back? These test two different things - the api may be available whilst the controller is in a crash loop.

hack/e2e.sh
```diff
@@ -19,83 +19,46 @@ source "${SCRIPT_DIR}/libe2e.sh"
...
-    echo "$1"
-    kubectl get po -o yaml
-    kubectl describe po
-    kubectl get svc -o yaml
-    kubectl describe svc
-    kubectl get apiservice -o yaml
-    kubectl describe apiservice
-    kubectl logs -c apiserver -l app=navigator,component=apiserver
-    kubectl logs -c controller -l app=navigator,component=controller
-    exit 1
```
munnerz: Why is this removed? This is here to make debugging failed tests easier 😄

munnerz: Ah, never mind, I see you've moved it.

wallrj:
Moved to https://github.com/jetstack/navigator/pull/67

PR #47, opened 2017-10-19
Update Kubernetes deps to use release-1.8 branches

munnerz:


munnerz:
Since https://github.com/kubernetes/kubernetes/pull/52186/ has merged, we can now really easily build generators and also generate files. We should gradually switch all of our projects over to use these new `generate-*` scripts.

This PR switches us to use this, fixes us to the `release-1.8` branches of the Kubernetes API repositories, and updates the import paths for the newly generated files.

cmd/apiserver/start.go
```diff
@@ -30,8 +30,8 @@ import (
 
 	"github.com/jetstack-experimental/navigator/pkg/apis/navigator/v1alpha1"
 	"github.com/jetstack-experimental/navigator/pkg/apiserver"
-	clientset "github.com/jetstack-experimental/navigator/pkg/client/clientset_generated/internalclientset"
-	informers "github.com/jetstack-experimental/navigator/pkg/client/informers_generated/internalversion"
+	clientset "github.com/jetstack-experimental/navigator/pkg/client/clientset/internalversion"
+	informers "github.com/jetstack-experimental/navigator/pkg/client/informers/internalversion"
```
wallrj: ❓ So how do we decide whether to use the `internalversion` or the `versioned` APIs? I guess I can see why the API server its self should use the internal versions, but why shouldn't the controller also use the internal APIs?

hack/update-client-gen.sh
```diff
@@ -6,36 +6,21 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
-BINDIR=${REPO_ROOT}/bin
-HACKDIR=${HACKDIR:-hack}
+SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
+CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
```
wallrj: ❓ I can't quite figure out what this is doing? Does it need an explanatory comment or is this copied over from Kubernetes?

wallrj:
Ah ignore me:
>    I checked out the branch and ran dep ensure -v --dry-run --no-vendor. It reports:
>    Gopkg.lock was not up to date
>    I ran make all and got the following error:

There were some lingering files frrom my cassandra work.

Gopkg.toml
```diff
@@ -30,25 +30,21 @@ required = [
 ]
 
 [[constraint]]
-  name = "github.com/Sirupsen/logrus"
-  revision = "v0.6.2-10-g51fe59a"
```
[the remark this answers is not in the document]
munnerz: Yep - this was just a constraint since we removed use of logrus (the package itself wasn't actually depended upon since before this PR)

Gopkg.toml
```diff
@@ -30,25 +30,21 @@ required = [
...
-
-[[constraint]]
   name = "k8s.io/client-go"
-  branch = "master"
+  branch = "release-5.0"
```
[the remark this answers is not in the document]
munnerz: Yes, the release branches are updated. We'll need to run `dep ensure -update` periodically to re-vendor those changes.

Makefile
```diff
@@ -94,43 +94,7 @@ go_fmt:
...
 # This section contains the code generation stuff
 #################################################
-.generate_exes: \
-	$(BINDIR)/defaulter-gen \
-	$(BINDIR)/deepcopy-gen \
-	$(BINDIR)/conversion-gen \
-	$(BINDIR)/client-gen \
-	$(BINDIR)/lister-gen \
-	$(BINDIR)/informer-gen
-	touch $@
```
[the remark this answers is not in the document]
munnerz: They now get compiled as part of the new script in code-generator. See: https://github.com/kubernetes/kubernetes/pull/52186/

cmd/apiserver/start.go
munnerz: Internal types should *only* be used within the API machinery for the apiserver, or if for *some reason* you need to convert between two external versions. That said, our controller should be built against one and only one version of the API, so there should basically never be a requirement to convert locally.

The controller shouldn't use the internal types for a number of reasons:

* These types can change, breaking compatibility.
* If we were to use the internal types, we'd then need to convert back into a versioned type in order to submit it back to the API server. We'd also need to convert *to* the internal version clientside too.
* I'd question what we have to gain from trying to do this
* It's also against all best practices for this sort of stuff, and would be a departure from how it is done within the k8s codebase. I can't recall an issue or justification for this decision, but I'm certain there is one with good reason.

There are a number of *other* potential complexities in doing this, and @dippynark and I did discuss it recently. I've managed to forget most of our thoughts on it though now! I will update here if they come back to me 😄

hack/update-client-gen.sh
munnerz: It's copied over from the Kubernetes codebase. Roughly:

* Set CODEGEN_PKG to CODEGEN_PKG if it *is* already set
* Otherwise, `cd` into the root of the repo (SCRIPT_ROOT) and attempt to list `./vendor/k8s.io/code-generator` if it exists, piping errors to /dev/null. If this command is successful, `./vendor/k8s.io/code-generator` is printed to stdout from `ls`, thus CODEGEN_PKG becomes `./vendor/k8s.io/code-generator`.
* Otherwise, return `../code-generator`. This final step is to deal with the case where our project is located within `$GOPATH/src/k8s.io` (or `k8s.io/kubernetes/staging/src/k8s.io/`). I copied this from the `sample-controller` project, hence this extra condition 😄 

The script itself will error regardless if this path is set incorrectly, so I'm not too bothered about falling back to a directory that likely doesn't exist.

munnerz:
Thanks for the review @wallrj - I've responded to your comments above.

Running `dep ensure -v --dry-run --no-vendor` locally I get: `Gopkg.lock was already in sync with imports and Gopkg.toml`. I guess whatever changes you had locally changed that 😄

PR #58, opened 2017-10-25
Create and synchronise a Cassandra service

wallrj:
Cassandra cluster database pods need to connect to each other.
A service provides a DNS name for the cluster and each pod will be given a stable hostname under that DNS name.
 * https://kubernetes.io/docs/tutorials/stateful-application/cassandra/#creating-a-cassandra-headless-service

When a CassandraCluster resource is added or modified, the Service control will ensure that a headless service is created for the cluster and it will update the service configuration if the CassandraCluster configuration changes. 

To run the tests:
```
go test -v ./pkg/controllers/cassandra/... -args -v 100 -alsologtostderr
```

```release-note
CassandraController now creates a Service resource.
```


pkg/controllers/cassandra/service/control.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (e *defaultCassandraClusterServiceControl) Sync(cluster *v1alpha1.CassandraCluster) error {
+	svc := ServiceForCluster(cluster)
+	_, err := e.kubeClient.CoreV1().Services(svc.Namespace).Update(svc)
+	if k8sErrors.IsNotFound(err) {
+		_, err = e.kubeClient.CoreV1().Services(svc.Namespace).Create(svc)
+	}
+	return err
+}
```
wallrj: @munnerz The ElasticSearch ServiceControl, doesn't `Update` existing services. I decided to add it here but is this the correct way to do the update?

munnerz: Yep it is, although we should be careful not to overwrite existing user data in an API server before performing the update, by checking the OwnerReference of the service. If the OwnerReference is *not* set to the CassandraCluster resource, we should bail out early and record an Event on the CassandraCluster resource.

wallrj: I added an owner reference check. See what you think.

pkg/controllers/cassandra/service/control.go
```diff
@@ -34,9 +37,17 @@ func NewControl(
...
 func (e *defaultCassandraClusterServiceControl) Sync(cluster *v1alpha1.CassandraCluster) error {
 	svc := ServiceForCluster(cluster)
-	_, err := e.kubeClient.CoreV1().Services(svc.Namespace).Update(svc)
+	client := e.kubeClient.CoreV1().Services(svc.Namespace)
+	existingSvc, err := e.serviceLister.Services(cluster.Namespace).Get(svc.Name)
 	if k8sErrors.IsNotFound(err) {
-		_, err = e.kubeClient.CoreV1().Services(svc.Namespace).Create(svc)
+		_, err := client.Create(svc)
+		return err
+	} else if err != nil {
```
munnerz: Can get rid of the `else` part here - if the error isn't not found, the next block will then check if it's not nil, and return err if it isn't nil 😄

pkg/controllers/cassandra/service/control.go
```diff
@@ -34,9 +37,17 @@ func NewControl(
...
+		return err
+	}
+	if !metav1.IsControlledBy(existingSvc, cluster) {
+		return fmt.Errorf("Duplicate service detected: %s", svc)
```
munnerz: This check looks good to me.

Can we update the error message to state that an existing service of name xyz already exists, and that we won't touch it? (so that users know they need to intervene). Also `%s` with `svc` (which is of type `v1.Service`) will give some ugly output that isn't useful for users (it'll be the address of the pointer). Can we update to be the service name? (`svc.Name`). It's the service name that has caused the conflict here, and it's by name that a user will fix this problem 😄

pkg/controllers/cassandra/service/control.go
```diff
@@ -34,9 +37,17 @@ func NewControl(
 
 func (e *defaultCassandraClusterServiceControl) Sync(cluster *v1alpha1.CassandraCluster) error {
 	svc := ServiceForCluster(cluster)
-	_, err := e.kubeClient.CoreV1().Services(svc.Namespace).Update(svc)
+	client := e.kubeClient.CoreV1().Services(svc.Namespace)
+	existingSvc, err := e.serviceLister.Services(cluster.Namespace).Get(svc.Name)
```
munnerz: (nit, but could be significant in future if we ever mess with namespaces - `svc.Namespace` or `cluster.Namespace`? I'd say `svc.Namespace` should be used, as the `ServiceForCluster` function takes care of ensuring that is correct.)

pkg/controllers/cassandra/cassandra.go
```diff
@@ -48,15 +51,22 @@ func NewCassandra(
...
-
-	cc.control = NewControl(recorder)
+	cc.control = NewControl(
+		service.NewControl(
+			kubeClient,
+			corev1listers.NewServiceLister(services.GetIndexer()),
+			recorder,
+		),
+		recorder,
+	)
```
munnerz: You've not stored a reference to the `HasSynced` function of the `services` SharedIndexInformer, so this controller will start *before* the service lister has synced. It should be done similar to how you've done with the `cassClusters` SharedIndexInformer.

pkg/controllers/cassandra/cassandra.go
```diff
@@ -172,6 +182,22 @@ func init() {
 					},
 				),
 			),
+			ctx.SharedInformerFactory.InformerFor(
+				ctx.Namespace,
+				metav1.GroupVersionKind{
+					Group:   navigator.GroupName,
+					Version: "corev1",
+					Kind:    "Service",
```
munnerz: `corev1` is not a version, and services do not exist in the navigator group.

`Group: ""`
`Version: "v1"`

(see similar examples in the ElasticsearchCluster controller)
(as discussed, I'm aware that this way of using the SharedInformerFactory sucks, and I promise it won't hang around forever, but whilst it does we need to make sure to use it correctly!)

pkg/controllers/cassandra/cassandra_test.go
```diff
@@ -1,30 +1,22 @@
-package cassandra
+package cassandra_test
```
munnerz: I guess you've done this so you can't access non-exported fields from the `cassandra` package during tests? If so, okay 😄

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -3,19 +3,20 @@ package cassandra
...
-	messageErrorSyncServiceAccount = "Error syncing service account: %s"
-	messageErrorSyncConfigMap      = "Error syncing config map: %s"
-	messageErrorSyncService        = "Error syncing service: %s"
-	messageErrorSyncNodePools      = "Error syncing node pools: %s"
-	messageSuccessSync             = "Successfully synced CassandraCluster"
+	MessageErrorSyncServiceAccount = "Error syncing service account: %s"
+	MessageErrorSyncConfigMap      = "Error syncing config map: %s"
+	MessageErrorSyncService        = "Error syncing service: %s"
+	MessageErrorSyncNodePools      = "Error syncing node pools: %s"
+	MessageSuccessSync             = "Successfully synced CassandraCluster"
```
munnerz: Is there a reason for exporting these? I'm not against it, just wondering.

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -25,22 +26,35 @@ type ControlInterface interface {
...
 	}
 }
 
 func (e *defaultCassandraClusterControl) Sync(c *v1alpha1.CassandraCluster) error {
 	glog.V(4).Infof("defaultCassandraClusterControl.Sync")
+	err := e.serviceControl.Sync(c)
+	if err != nil {
+		e.recorder.Eventf(
+			c,
+			"cassandra.defaultCassandraClusterControl",
```
munnerz: Event types should be one of `v1.EventTypeNormal` or `v1.EventTypeWarning`

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -25,22 +26,35 @@ type ControlInterface interface {
...
+			ErrorSync,
+			MessageErrorSyncService,
```
munnerz: `MessageErrorSyncService` = `"Error syncing service: %s"` - you are using `Eventf` and your final argument is the `*v1alpha1.CassandraCluster`. I think this was meant to be `err.Error()` 😄

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -25,22 +26,35 @@ type ControlInterface interface {
...
+			c,
+		)
+		return err
+	}
 	e.recorder.Event(
 		c,
 		"cassandra.defaultCassandraClusterControl",
```
munnerz: Again, this should be `v1.EventTypeNormal`. We shouldn't be just making these up 😄

pkg/controllers/cassandra/util/util.go
```diff
@@ -0,0 +1,37 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/jetstack-experimental/navigator/pkg/apis/navigator"
+	v1alpha1 "github.com/jetstack-experimental/navigator/pkg/apis/navigator/v1alpha1"
+)
+
+const (
+	typeName = "cass"
+	kindName = "CassandraCluster"
```
munnerz: We really should come up with a better way to derive `kindName`... there's a few ways we can do it. But this isn't something to worry about now.

pkg/controllers/cassandra/service/control.go
wallrj: Done.

pkg/controllers/cassandra/cassandra.go
wallrj: Done.

pkg/controllers/cassandra/cassandra.go
wallrj: Deon.

pkg/controllers/cassandra/cassandra_test.go
wallrj: In the previous version of the tests I mistakenly thought that I needed to wait for the cache sync before simulating resource  create / update / delete  in the tests.
So I grabbed the private cassListerSynced attribute from the controller.
But experimented some more and found that that isn't actually necessary.

pkg/controllers/cassandra/cluster_control.go
wallrj: I made them public so that I can look for the sync success or failure messages in the fake event recorded....to know when the Sync operation has completed in the  integration test in cassandra_test.go.

pkg/controllers/cassandra/cluster_control.go
wallrj: Done.

munnerz:
Ah - tests are failing because the repo has moved to be under the `jetstack` org. You'll need to rebase your changes and update imports :smile:

pkg/controllers/cassandra/cluster_control.go
wallrj: Done.

pkg/controllers/cassandra/cluster_control.go
wallrj: Done.

pkg/controllers/cassandra/service/control.go
wallrj: Good point. Done.

pkg/controllers/cassandra/service/control.go
wallrj: done.

pkg/controllers/cassandra/service/control.go
wallrj: Good point. Done.

wallrj:
@munnerz Thanks for the review. I think I've addressed all your and merged in the jetstack-experimental -> jetstack changes so that the CI now passes. Please take another look.

pkg/controllers/cassandra/service/control_test.go
```diff
@@ -55,4 +55,16 @@ func TestServiceSync(t *testing.T) {
 			}
 		},
 	)
+	t.Run(
+		"service with outside owner",
+		func(t *testing.T) {
+			f := casstesting.NewFixture(t)
+			// Remove the ports from the default cluster and expect them to be
+			// re-created.
```
munnerz: This comment seems to be a c+p error?

munnerz: Happy to leave this here though and you can remove it in the upstream branch or something if you're going to be working in this file again anyway 😄  your call

wallrj: Ah yes. I'd prefer to fix that in a followup branch rather than wait for another round of CI.

wallrj:
Thanks !

PR #59, opened 2017-10-25
Create and synchronise StatefulSets for CassandraCluster

wallrj:
The CassandraCluster ring processes will be managed by a StatefulSet.

* https://kubernetes.io/docs/tutorials/stateful-application/cassandra/#using-a-statefulset-to-create-a-cassandra-ring

This controller helper will ensure that a StatefulSet is created for each  CassandraClusterNodePool in CassandraCluster.Spec.NodePools .

There's more work to do on creating and managing stateful sets but i don't want this PR to get too big.

```release-note
CassandraController now creates StatefulSet resources
```


pkg/controllers/cassandra/cassandra.go
```diff
@@ -200,6 +213,21 @@ func init() {
 					},
 				),
 			),
+			ctx.SharedInformerFactory.InformerFor(
+				ctx.Namespace,
+				metav1.GroupVersionKind{
+					Version: "v1beta2",
+					Kind:    "StatefulSet",
```
munnerz: Don't forget the group here

wallrj: Done. But note that it doesn't seem to make any difference in my testing.

munnerz: It won't make a difference, but it'll create an extra SharedIndexInformer instead of re-using the existing one, which will cause us to store twice as many StatefulSets in memory 😄 

We should also use v1beta1 here instead of v1beta2, in order to support API servers that don't support v1beta2

munnerz: (BTW - my PR to kubernetes/kubernetes has been updated today, so I'm hoping all this horrible stuff will go away soon 😄 )

munnerz: Ah - just seen you've already updated the version in a later commit!

wallrj: Thanks. In https://github.com/jetstack/navigator/pull/64 I've refactored things a bit so that I can exercise this code in the integration test.

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -23,7 +23,14 @@ type CassandraCluster struct {
 }
 
 type CassandraClusterSpec struct {
-	Sysctl []string `json:"sysctl"`
+	Sysctl    []string                   `json:"sysctl"`
+	NodePools []CassandraClusterNodePool `json:"nodePools"`
+}
+
+// CassandraClusterNodePool describes a node pool within an CassandraCluster.
```
munnerz: (nit) an->a

pkg/controllers/cassandra/cassandra.go
```diff
@@ -31,19 +35,21 @@ import (
...
-	cassLister          listersv1alpha1.CassandraClusterLister
-	cassListerSynced    cache.InformerSynced
-	serviceListerSynced cache.InformerSynced
-	queue               workqueue.RateLimitingInterface
-	recorder            record.EventRecorder
+	control                  ControlInterface
+	cassLister               listersv1alpha1.CassandraClusterLister
+	cassListerSynced         cache.InformerSynced
+	serviceListerSynced      cache.InformerSynced
+	statefulSetsListerSynced cache.InformerSynced
```
munnerz: (nit) statefulSets -> statefulSet (for consistency with other vars in this struct)

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -51,6 +58,17 @@ func (e *defaultCassandraClusterControl) Sync(c *v1alpha1.CassandraCluster) erro
 		)
 		return err
 	}
+	err = e.nodepoolControl.Sync(c)
+	if err != nil {
+		e.recorder.Eventf(
+			c,
+			"cassandra.defaultCassandraClusterControl",
```
munnerz: Event type is not arbitrary

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -51,6 +58,17 @@ func (e *defaultCassandraClusterControl) Sync(c *v1alpha1.CassandraCluster) erro
...
+			ErrorSync,
+			MessageErrorSyncNodePools,
+			c,
```
munnerz: `c` is a pointer to a structure, not a string. This will print badly.

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -0,0 +1,131 @@
...
+	"k8s.io/client-go/tools/record"
+)
+
+type Interface interface {
+	Sync(*v1alpha1.CassandraCluster) error
+}
+
+type defaultCassandraClusterNodepoolControl struct {
+	kubeClient        kubernetes.Interface
+	statefulsetLister appslisters.StatefulSetLister
```
munnerz: (nit) in the `cassandra_control.go` file, you call this `statefulSet` instead of `statefulset`

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -0,0 +1,131 @@
...
+		expectedStatefulSetNames[name] = true
+	}
+	client := e.kubeClient.AppsV1beta1().StatefulSets(cluster.Namespace)
+	selector, err := util.SelectorForCluster(cluster)
+	if err != nil {
+		return err
+	}
+	existingSets, err := e.statefulsetLister.
+		StatefulSets(cluster.Namespace).
+		List(selector)
```
munnerz: (nit) collapse to one line (like `client` above)?

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return err
+	}
+	for _, set := range existingSets {
+		err := ownerCheck(set, cluster)
+		if err != nil {
+			return err
+		}
```
munnerz: Does it make sense to return an error here, if a cluster is found with matching labels but the ownerRef is wrong? This will cause the rest of the node pool synchronisation to fail (and also cause the sync to be retried)

If it does make sense to return an error here and have it retried later, we should log an Event to the CassandraCluster resource to make it clear that a StatefulSet matching labels but not owned by blah blah blah exists and cannot be modified (and recommending the course of action to a user).

That said, I'd have thought we're best to just ignore that StatefulSet and perhaps log a warning that we came across something that matches labels, but isn't owned.

munnerz: Ah - I've realised that the error returned from `ownerCheck` is already quite descriptive. My only remaining point then is just verifying that this should cause this sync to be aborted and retried later.

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -0,0 +1,24 @@
...
+) *apps.StatefulSet {
+	set := &apps.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            util.NodePoolResourceName(cluster, nodepool),
+			Namespace:       cluster.Namespace,
+			Labels:          util.ClusterLabels(cluster),
+			Annotations:     make(map[string]string),
+			OwnerReferences: []metav1.OwnerReference{util.NewControllerRef(cluster)},
+		},
+	}
```
munnerz: I guess this is TODO still? (for a follow up PR?)

pkg/apis/navigator/v1alpha1/types.go
wallrj: Done.

pkg/controllers/cassandra/cassandra.go
wallrj: Done

pkg/controllers/cassandra/cluster_control.go
wallrj: Done.

pkg/controllers/cassandra/cluster_control.go
wallrj: Done.

pkg/controllers/cassandra/nodepool/nodepool.go
wallrj: Done.

pkg/controllers/cassandra/nodepool/nodepool.go
wallrj: Done.

pkg/controllers/cassandra/nodepool/nodepool.go
wallrj: I thought it safest to err here after vaguely  thinking of the case where user creates > deletes > recreates a cluster with the same name.
Until we implement  garbage collect of the the old resources we'll run into this problem, I think.

pkg/controllers/cassandra/nodepool/resource.go
wallrj: Yep, there's more to come in #65

wallrj:
> Only quick question, and can be addressed in other PRs - does it make sense to have nodePools in a cassandra cluster? I'd guess yes, but just want to be sure. 

I'm not sure yet. I was tempted to not include it but thought it best to stay as consistent with ES API as I can...for now.
As far as I can see, Cassandra doesn't allow separation of node roles (data, client, etc).
All nodes perform all roles.
You can have different spec nodes with different numbers of shards, but various docs warn against doing that....although I can't find the links right now.

pkg/controllers/cassandra/nodepool/nodepool.go
munnerz: Garbage collection wouldn't prevent this though, as it would only garbage collect things that are 'owned'. This code basically says "if a StatefulSet that has the label X=Y on it exists, and it's not owned by this CassCluster, this is an error state and we cannot proceed".

This makes sense to me *iff* that StatefulSet *also* has the same name as another statefulset that we *do* want to own (because we wouldn't want to overwrite an existing deployment). Right now though, we just bail out even if it would be possible to proceed from what I can see.

In the case where a user creates->deletes->creates, the check for ownerRef [here](https://github.com/jetstack/navigator/pull/59/files/f05f42e03efdaaae8dec1ff34c7b30dea96c597a#diff-9c2ba35894d4eb5cf3a1e42b156eac33R106) should pick it up and bail out (when the actual create happens).

munnerz:
@wallrj regarding node pools - cool, let's leave it in for now. We can always adjust before this gets merged anyway. I think we're definitely best to follow best practices here. I think I may have exposed *too much* API surface in the ES cluster controller right now as it is.

We need to make sure we walk the line between a managed service and customisable. We want users to be able to upgrade clusters by changing a single field IMO. Navigator should work out configuration changes that need applying in order to do so (unless the user has manually overridden something, in which case we go hands off and it's up to them to do things 'right')

pkg/controllers/cassandra/nodepool/nodepool.go
munnerz: Will lgtm just to keep this moving, but I do think we should revisit this clause 😄

PR #65, opened 2017-10-30
Add StatefulSet pod spec 

wallrj:
Adds a meaningful Pod spec to the statefulsets that are generated

Release Note: 

```release-note
NONE
```


wallrj:
@munnerz Sorry this has taken so long. I think it's ready for a review now.

* I've got the stage where we can start real Cassandra database pods
* Added an e2e test to see that the service allows access to the CQL port
* Added a Docker image that I used for debugging problems with the service connections and which might be useful in the future.
* Added a basic Cassandra chart which has been useful for overriding configuration settings such as pull policy and image.
* Fixed missing selector on the cassandra service
* Added unique namespaces for each test run, because deleting namespaces is so slow.

In followup branch I'll add Pilot and Pilot Resource syncing.

https://kubernetes.io/docs/tutorials/stateful-application/cassandra/#using-a-statefulset-to-create-a-cassandra-ring

PR #70, opened 2017-10-31
Fix leader election with RBAC in tests

munnerz:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/contrib/charts/navigator/templates/rbac.yaml
+++ b/contrib/charts/navigator/templates/rbac.yaml
@@ -88,9 +88,9 @@ items:
   rules:
   - apiGroups: ["navigator.jetstack.io"]
     resources: ["elasticsearchclusters", "pilots"]
-    verbs:     ["get", "list", "watch", "update"]
+    verbs:     ["get", "list", "watch", "update", "create", "delete"]
   - apiGroups: [""]
-    resources: ["services","configmaps","serviceaccounts"]
+    resources: ["services", "configmaps", "serviceaccounts", "pods"]
     verbs:     ["get", "list", "watch", "update", "create", "delete"]
   - apiGroups: ["apps"]
     resources: ["statefulsets"]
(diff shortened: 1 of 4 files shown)


munnerz:
This PR also disables scoping Navigator to a single namespace as part of the standard helm chart deployment

contrib/charts/navigator/templates/rbac.yaml
```diff
@@ -95,6 +95,10 @@ items:
   - apiGroups: ["apps"]
     resources: ["statefulsets"]
     verbs:     ["get", "list", "watch", "update", "create", "delete"]
+  - apiGroups: [""]
+    resources: ["endpoints"]
+    verbs:     ["*"]
+    resourceNames: ["navigator-controller"]
```
wallrj: ❓ What does this do?

wallrj:
[log.gz](https://github.com/jetstack/navigator/files/1430669/log.gz)

munnerz:
So after looking into this some more, the actual controller won't be able to create the initial Endpoint resource used for leader election as we have explicitly named the resource in the RBAC policy, which consequently then rejects POST requests to the API endpoint (as the endpoint does *not* include the resource name at this point). This is a limitation with RBAC at the moment.

contrib/charts/navigator/templates/rbac.yaml
munnerz: This allows navigator-controller to modify Endpoints resources with the name `navigator-controller`. This is the resource used for leader election 😄

munnerz:
Updated & tested working. Review as and when you've got time. Don't want to hold you up!

munnerz:
@wallrj - I'm well aware the e2e tests need improving. Currently we are blocked by needing to supply some e2e testing infrastructure via Prow, so that we can actually get VMs big enough to run Elasticsearch (it is a massive resource hog).

This will I'll be working on some e2e testing, so will get the actual test cases written up then so we can start to see how they perform 😄

Issue #76, opened 2017-10-31
Travis builds everything twice


wallrj:
Travis is running builds twice: 1) for branches and 2) for PR

Instead you can configure it to only run the builds for certain branches and for PRs like so:

```
# Only run tests on push on a few branches.
# Test on PR should be execute for all branches and forks.
branches:
  only:
  - master
- /^release\/flocker-.*$/
```

See: https://github.com/ClusterHQ/flocker/blob/master/.travis.yml#L44

This way it'll also build the master branch when PRs are merged. 

/kind bug

PR #77, opened 2017-10-31
Only run Travis-CI tests for PRs and when master branch changes

wallrj:
Fixes: #76 

This PR should only run the Travis-CI tests once, for the PR not for the branch.

Should speed up the time it takes to approve and merge small PRs

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,3 +42,8 @@ jobs:
           mkdir -p ~/.docker && echo "${DOCKER_AUTH_CONFIG}" > ~/.docker/config.json && chmod 600 ~/.docker/config.json;
           make docker_push IMAGE_TAGS="${TRAVIS_COMMIT} latest";
         fi
+
+# Only run tests when master branch changes or when a PR branch is updated.
+branches:
+  only:
+  - master


munnerz:
I'm not adverse to this, but given the two jobs run in parallel I feel like there's little to gain by doing this. The *only* reason I can think why we *would* keep testing both the PR and the branch, is to make it easy to see that the reason your PR is failing, is due to the the merge with master. This is pretty useful and could save someone spending ages trying to work out why something passes locally, but fails in CI (and as such, requires rebasing and fixing up against master).

wallrj:
Hey @munnerz thanks for the review and for merging this.
I must say that I didn't realise that subtle difference between branch and PR builds.
I agree that this won't make the builds faster, but hopefully with fewer tests to run, we'll reduce the likelihood of spurious test failures due to minikube flakiness / network problems etc.

munnerz:
👍  let's leave it like it is for now. If we start getting issues that would be easier to debug *with* the additional tests I think we should add them back in.

PR #79, opened 2017-10-31
Use custom informer-gen that supports filters

munnerz:
This PR switches us to use `munnerz/code-generator` temporarily, which includes the changes in  https://github.com/kubernetes/kubernetes/pull/54660. I've then updated our codegen scripts to *also* generate informers for internal kubernetes types (although *not* a client or lister, as we can re-use client-go for this).

I'm hoping my PR will be merged soon, at which point we can remove the alternate repo source, but for now this lets us clear up the mess that is our SharedInformerFactory.

```release-note
NONE
```


munnerz:
Yeah that is what I hoped to do originally. `dep` isn't able to handle that, and actually their justifications are fair (if two packages both depend on the same package, but one of them specifies a different 'source', what should happen, as we can't check out code from two places so the dependency cannot be met).

It'd be nice if there were *some* way to do it, but doesn't seem like it.

I've actually just run into a new problem, which is that dep wipes out the `vendor/` dir upon `dep ensure`, so I've had to actually update the Makefile to include a `ln -s` which is not the best. I'm *really* hoping my PR gets accepted upstream so we don't have to do this forever! There's been some good/positive discussion on it, so I fingers crossed :smile:

wallrj:
Hey @munnerz 

I just read: https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#constraint

```
[[constraint]]
...
  # Optional: an alternate location (URL or import path) for the project's source.
  source = "https://github.com/myfork/package.git"
```

Found via the response to this comment: https://github.com/golang/dep/issues/248#issuecomment-281583583

munnerz:
Thanks @wallrj! Missed that!!

munnerz:
@wallrj this is now passing! :tada:

I've also opened a follow up PR #82 that switches the ElasticsearchController to use these new factories.

Gopkg.toml
```diff
@@ -45,6 +45,8 @@ required = [
   name = "k8s.io/apiserver"
   branch = "release-1.8"
 
+# Use munnerz/code-generator fork that includes the filter changes
 [[constraint]]
   name = "k8s.io/code-generator"
   branch = "release-1.8"
```
wallrj: ❓ I was going to ask if this is the correct branch in your forked repo, but it looks like you've applied the changes there too: https://github.com/munnerz/code-generator/tree/release-1.8

hack/update-client-gen.sh
```diff
@@ -24,3 +24,12 @@ ${CODEGEN_PKG}/generate-internal-groups.sh "conversion" \
...
   --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.go.txt \
   --extra-peer-dirs="k8s.io/api/core/v1,k8s.io/apimachinery/pkg/apis/meta/v1,k8s.io/apimachinery/pkg/conversion,k8s.io/apimachinery/pkg/runtime"
+
+${GOPATH}/bin/informer-gen \
+           --output-base "${GOPATH}/src/" \
+           --input-dirs "k8s.io/api/core/v1" \
+           --input-dirs "k8s.io/api/apps/v1beta1" \
+           --versioned-clientset-package "k8s.io/client-go/kubernetes" \
+           --listers-package "k8s.io/client-go/listers" \
+           --output-package "github.com/jetstack/navigator/third_party/k8s.io/client-go/informers" \
```
wallrj: ❓ Are we still using the `third_party` sub-directory

Gopkg.toml
munnerz: Yep - otherwise there were more changes to generated files that were unrelated. Thought it best to make this 1:1

hack/update-client-gen.sh
munnerz: Yep, but not with any vendor hacks this time (symlinks).

This is just a place to store the generated filter-able informers from the main k8s codebase.

Issue #78, opened 2017-10-31
Test against Kubernetes 1.8


wallrj:
We currently only test against Kubernetes 1.7.0

We should test Navigator against Kubernetes 1.8 too

/kind feature

Latest version of Minikube installs kubernetes 1.8 by default.

 * https://github.com/kubernetes/minikube/blob/v0.23.0/CHANGELOG.md#version-0230---10262017

PR #80, opened 2017-10-31
Run e2e tests against Kubernetes 1.8.0 too

wallrj:
Fixes: #78 

**Release note**:
```release-note
Navigator is now tested on Kubernetes 1.8.0
```

### DIFF
--- a/hack/install-e2e-dependencies.sh
+++ b/hack/install-e2e-dependencies.sh
@@ -1,0 +1,25 @@
+set -eux
+curl -Lo helm.tar.gz \
+     https://storage.googleapis.com/kubernetes-helm/helm-v2.6.1-linux-amd64.tar.gz
+tar xvf helm.tar.gz
+sudo mv linux-amd64/helm /usr/local/bin
+
+curl -Lo kubectl \
+     https://storage.googleapis.com/kubernetes-release/release/$KUBERNETES_VERSION/bin/linux/amd64/kubectl
+chmod +x kubectl
+sudo mv kubectl /usr/local/bin/
+
+curl -Lo minikube \
+     https://storage.googleapis.com/minikube/releases/v0.23.0/minikube-linux-amd64
+chmod +x minikube
+sudo mv minikube /usr/local/bin/
+
+docker run -v /usr/local/bin:/hostbin quay.io/jetstack/ubuntu-nsenter cp /nsenter /hostbin/nsenter
+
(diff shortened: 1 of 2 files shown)


wallrj:
Hmm. No that doesn't work. 
Need to figure out how to add a matrix of environment variables but only expand those for the e2e tests.

munnerz:
Yeah I've tried to do this a few times myself for cert-manager. The control plane never seems to actually come up though, and I've not had time to debug further - I did sort out the matrix expansion however 😄  https://github.com/jetstack-experimental/cert-manager/pull/106

wallrj:
I made a script to install e2e test dependencies rather than including it all in the travis yaml.
And rather than add a top level ENV matrix, which would have also caused the unit tests to be run twice, I created two test jobs each with its own kubernetes_version env.

We could speed things up further by removing the unit test step, which is already performed by prow.
And we could move the build step first and re-use the built binaries in the e2e tests.
Perhaps by uploading binaries to a bucket or by imaginitive use of the travis cache feature:
 * https://docs.travis-ci.com/user/caching/#Arbitrary-directories

@munnerz Please review this when you have a moment.

hack/install-e2e-dependencies.sh
```diff
@@ -0,0 +1,25 @@
+set -eux
+curl -Lo helm.tar.gz \
```
munnerz: doesn't need to be fixed in this PR - we should have a consistent header for bash files/files in general, and have it verified by CI that all files are up to date. Not one for this PR though, we don't have any 'enforced' style at the moment.

hack/install-e2e-dependencies.sh
```diff
@@ -0,0 +1,25 @@
...
+
+curl -Lo kubectl \
+     https://storage.googleapis.com/kubernetes-release/release/$KUBERNETES_VERSION/bin/linux/amd64/kubectl
+chmod +x kubectl
+sudo mv kubectl /usr/local/bin/
+
+curl -Lo minikube \
+     https://storage.googleapis.com/minikube/releases/v0.23.0/minikube-linux-amd64
+chmod +x minikube
+sudo mv minikube /usr/local/bin/
```
munnerz: Doesn't need to be done for this, as it's a whole new unit of feature, but now that we're doing this in a script in hack (i.e. it's more likely to be run by developers and not just CI), I think we shouldn't make changes outside of the repo root/in `/usr/local/bin`

PR #82, opened 2017-11-01
Use filtered SharedInformerFactories in controller context

### LINKED ISSUE #69: Getting rid of verbose and fragile informer construction
Right now we construct informers through a really horrible and non-type safe SharedInformerFactory, in order to support running in a single namespace (see: https://github.com/jetstack/navigator/blob/master/pkg/controllers/elasticsearch/elasticsearch.go#L311)

We should update this to use a custom generated SharedInformerFactory that supports namespaced resources, and simplify all of this informer handling logic for good.

munnerz:

Switches to using the new filtered SharedInformerFactories as part of https://github.com/kubernetes/kubernetes/pull/54660

fixes #69 

Don't merge until #79 has been merged

**Release note**:

```release-note
NONE
```

/assign @wallrj


pkg/controllers/elasticsearch/elasticsearch.go
```diff
@@ -311,41 +310,13 @@ func (e *ElasticsearchController) handlePodObject(obj interface{}) {
...
-				metav1.GroupVersionKind{Version: "v1", Kind: "ConfigMap"},
-				coreinformers.NewConfigMapInformer(ctx.Client, ctx.Namespace, time.Second*30, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}),
-			),
+			ctx.SharedInformerFactory.Navigator().V1alpha1().ElasticsearchClusters(),
+			ctx.SharedInformerFactory.Navigator().V1alpha1().Pilots(),
+			ctx.KubeSharedInformerFactory.Apps().V1beta1().StatefulSets(),
+			ctx.KubeSharedInformerFactory.Core().V1().Pods(),
+			ctx.KubeSharedInformerFactory.Core().V1().ServiceAccounts(),
+			ctx.KubeSharedInformerFactory.Core().V1().Services(),
+			ctx.KubeSharedInformerFactory.Core().V1().ConfigMaps(),
```
wallrj: 👍 Much neater!

munnerz: Our controller constructors are also typed now *at last* :tada: (no more horribly difficult to debug errors!)

cmd/controller/app/controller.go
```diff
@@ -177,3 +182,9 @@ func KubeConfig(apiServerHost string) (*rest.Config, error) {
 
 	return cfg, nil
 }
+
+func NamespaceFilter(namespace string) func(listOptions *metav1.ListOptions) string {
```
wallrj: ❓ Could you use type `FilterFunc` for the return signature? Is `FilterFunc` public? If not, should it be?

pkg/controllers/elasticsearch/elasticsearch.go
```diff
@@ -69,13 +68,13 @@ type ElasticsearchController struct {
...
-	serviceaccounts cache.SharedIndexInformer,
-	services cache.SharedIndexInformer,
-	configmaps cache.SharedIndexInformer,
+	es navigatorinformers.ElasticsearchClusterInformer,
+	pilots navigatorinformers.PilotInformer,
+	statefulsets appsinformers.StatefulSetInformer,
+	pods coreinformers.PodInformer,
+	serviceaccounts coreinformers.ServiceAccountInformer,
+	services coreinformers.ServiceInformer,
+	configmaps coreinformers.ConfigMapInformer,
```
wallrj: ❓ So this is where we get better type checking. Great.

third_party/k8s.io/client-go/informers/externalversions/apps/interface.go
```diff
@@ -0,0 +1,45 @@
+/*
+Copyright 2017 Jetstack Ltd.
```
wallrj: ❓ This should be Copyright Kubernetes I guess.

cmd/controller/app/controller.go
munnerz: The issue is that the `FilterFunc` type is generated within each `informer` package, meaning that we'd need to explicitly cast between the two types when we call `NewSharedInformerFactory`.

Ultimately it's a limitation in Go's type system - it doesn't inspect the 'value' side of the type when assessing whether something in fact implements the type, so two types that look identical to you or I, are different according to Go (although *can* be cast, which *does* take account of the 'value' side). There's a good blog post on this somewhere, but can't find it right now. Ping me if you want me to keep digging for it :smile:

third_party/k8s.io/client-go/informers/externalversions/apps/interface.go
munnerz: I'm not really sure to be honest - it's generated code, so I feel like it's not something that needs to be attributed to the k8s authors. If we were to change *this* to refer to the k8s authors, I feel like we should change *all* code generated by code-generators to be credited to the k8s authors.

PR #90, opened 2017-11-01
Various e2e test improvements which help when running the tests locally

wallrj:
* Create a unique namespace for each test.
  This helps when re-running the tests on a local minikube where namespace
  deletion takes up to 10 minutes!
  Delete the namespace if the test succeeds else leave it so that the failed
  resources can be examined.
* Only show logs from the time of the test run.
  Again, this helps with debugging when running the tests repeatedly, locally.
* Show log timestamps.

```release-note
NONE
```

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -16,8 +16,9 @@ mkdir -p $TEST_DIR
 
 source "${SCRIPT_DIR}/libe2e.sh"
 
+START_TIME="$(date +"%s")"
+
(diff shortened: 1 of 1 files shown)


wallrj:
I misunderstood.
* The logs already contain timestamps.
* And `kubectl logs deployment/navigator-controller` only ever show the logs from the latest deployment, so there's no need for the `--since Xs` stuff.

PR #93, opened 2017-11-02
Elasticsearch Pilot RBAC policies

### LINKED ISSUE #92: RBAC policy for pilots
/kind feature

When we create Pilots, we need to ensure correct RBAC policies are applied to their service accounts.

munnerz:

Creates RBAC policies for created Pilot resources

Fixes #92 

**Release note**:
```release-note
Automatically create RBAC policies for Pilots
```


munnerz:
@wallrj this is ready for review when you have a chance 😄

munnerz:
@wallrj if you want to hold off a sec, I'm just working on adding a test for this too (similar to yours)

munnerz:
This now depends on #99, as otherwise a privilege escalation error is thrown when creating the Pilots role (as the controller *currently* isn't allow to patch events)

munnerz:
@wallrj please take a look when you have a chance. #99 should probably be merged first (although I don't *think* it'll matter too much)

munnerz:
>@munnerz Did you consider / is it possible to create the necessary service accounts and rbac roles / from the helm chart?

We could theoretically create a ClusterRole resource for each Pilot type that the current version of Navigator supports, however that introduces two limitations:

1) We would have to make sure the user uses the correct version of the Helm chart for the correct version of Navigator. This puts more burden on the user deploying Navigator to do it correctly, and could be hard to debug. Users may also deploy Navigator without Helm, which would make it even harder for them to get this right.

2) All changes to the Pilot RBAC role will *always* have to be additive in future, unless the operator coordinates updating every Pilot simultaneously (i.e. with downtime). Think in the case where we introduce a new version of Pilot that requires a new RBAC permission, but drops another. We'd have to turn off all instances of Pilot (i.e. every database of that particular type) in order to update the role, otherwise we'd break older versions of Pilot.

3) That one ClusterRole is a SPOF for all databases of a given type, which is probably the least malicious of the 3 points, but still a consideration.

> In this PR we're granting navigator permission to modify any RBAC role / binding, right?
> Isn't that going to spook security concious kubernetes administrators?

So Kubernetes disallows privilege escalation in RBAC, so navigator won't be able to grant itself any more permission than it already has. This should give security conscious admins peace of mind that Navigator is only handing off permission to do something it already does.

pkg/controllers/elasticsearch/role/controller.go
```diff
@@ -0,0 +1,78 @@
...
+		return fmt.Errorf(
+			"foreign owned Role: "+
+				"A Role with name '%s/%s' already exists, "+
+				"but it is controlled by '%v', not '%s/%s'",
+			role.Namespace, role.Name, ownerRef,
+			cluster.Namespace, cluster.Name,
+		)
+	}
+	return nil
+}
```
wallrj: ❓ Can this be made a generic helper function that accepts type `cluster metav1.ObjectMeta`  ?

pkg/controllers/elasticsearch/role/resources.go
```diff
@@ -0,0 +1,33 @@
...
+				Resources: []string{"events"},
+			},
+			{
+				APIGroups: []string{navigator.GroupName},
+				Verbs:     []string{"update", "get"},
+				Resources: []string{"pilots", "elasticsearchclusters"},
+			},
+		},
+	}
+}
```
wallrj: ❓ Maybe this function should be called `pilotRoleForCluster` (or something less generic)

pkg/controllers/elasticsearch/role/resources.go
```diff
@@ -0,0 +1,33 @@
...
+		Rules: []rbacv1beta1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Verbs:     []string{"create", "update", "patch"},
+				Resources: []string{"events"},
+			},
+			{
+				APIGroups: []string{navigator.GroupName},
+				Verbs:     []string{"update", "get"},
+				Resources: []string{"pilots", "elasticsearchclusters"},
```
wallrj: ❓ Doesn't a Pilot controller need to be able to watch and list Pilots?

pkg/controllers/elasticsearch/rolebinding/controller.go
```diff
@@ -0,0 +1,78 @@
...
+		return fmt.Errorf(
+			"foreign owned Role: "+
+				"A Role with name '%s/%s' already exists, "+
+				"but it is controlled by '%v', not '%s/%s'",
+			role.Namespace, role.Name, ownerRef,
+			cluster.Namespace, cluster.Name,
+		)
+	}
+	return nil
+}
```
wallrj: ❓ Yeah let's avoid duplicating this function everywhere, if possible.

pkg/controllers/elasticsearch/rolebinding/controller.go
```diff
@@ -0,0 +1,78 @@
...
+	if err := ownerCheck(existingRoleBinding, c); err != nil {
+		return err
+	}
+	// TODO: avoid using `reflect` here
+	if !reflect.DeepEqual(desiredRoleBinding.RoleRef, existingRoleBinding.RoleRef) || !reflect.DeepEqual(desiredRoleBinding.Subjects, existingRoleBinding.Subjects) {
+		_, err = e.kubeClient.RbacV1beta1().RoleBindings(desiredRoleBinding.Namespace).Update(desiredRoleBinding)
+		return err
+	}
+	return nil
+}
```
wallrj: ❓ And this function seems almost identical to `RoleControl.Sync`. Can we turn this into a generic `createOrUpdate` ? Here or in a follow up.

third_party/k8s.io/client-go/informers/externalversions/rbac/interface.go
```diff
@@ -0,0 +1,45 @@
+/*
+Copyright 2017 Jetstack Ltd.
```
wallrj: ❓ Wrong copyright header. Should be (c) Kuberneties authors I'd have thought.

wallrj:
Oops! I didn't expect that to be merged. Oh well.

pkg/controllers/elasticsearch/role/controller.go
munnerz: That's basically what `metav1.IsControlledBy` does already, except this function prints a little bit more info including the error message containing the Kind names (Role etc.)

pkg/controllers/elasticsearch/role/resources.go
munnerz: This isn't generic between controllers - cassandra pilots shouldn't be able to access elasticsearchclusters. I'd also like to keep it quite explicit and obvious what we're granting, at least for now.

pkg/controllers/elasticsearch/rolebinding/controller.go
munnerz: This isn't identical - we check different heuristics to determine whether an update is required (see L73 in this file). We'd also then need to lose the type safety in the client in order to do this.

I'd rather consume a few kb more space on disk to make this code easier to read (through the use of types, etc.).

If Go's type system wasn't so picky about interfaces, we could potentially pass the `e.roleBindingLister.RoleBindings(desiredRoleBinding.Namespace)` to some kind of a generic function so that it can call `Update`/`Get`/`Create`, but the type signature of those functions is `Update(*v1beta1.RoleBinding) (*v1beta1.RoleBinding, error)` and not `Update(runtime.Object) (runtime.Object, error)`, and despite *v1beta1.RoleBinding *implementing* `runtime.Object`, the two functions are not compatible.

third_party/k8s.io/client-go/informers/externalversions/rbac/interface.go
munnerz: As I mentioned [here](https://github.com/jetstack/navigator/pull/82#discussion_r148257631) about generated code - I don't think it should be. The authors wrote the generators, I ran them. I depend on external code that *is* licensed by the Kubernetes authors, but the code in third_party is mine & I generated it. We depend on third party projects everywhere, but that doesn't mean those we depend on own the license on navigator.

> I'm not really sure to be honest - it's generated code, so I feel like it's not something that needs to be attributed to the k8s authors. If we were to change this to refer to the k8s authors, I feel like we should change all code generated by code-generators to be credited to the k8s authors.

pkg/controllers/elasticsearch/role/resources.go
munnerz: Yep it will do

PR #96, opened 2017-11-02
WIP: Fail e2e tests when there are unexpected errors in the logs

wallrj:
Filter the logs for errors that are not expected.
e.g. Errors caused by the controller attempting to list navigator API resources before the navigator API has been registered.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/docs/quick-start/es-cluster-demo.yaml
+++ b/docs/quick-start/es-cluster-demo.yaml
@@ -21,7 +21,7 @@ spec:
     fsGroup: 1000
 
   nodePools:
-  - name: MasterPool
+  - name: master
     replicas: 3
 
     roles:
--- a/docs/quick-start/es-cluster-demo.yaml
+++ b/docs/quick-start/es-cluster-demo.yaml
@@ -21,7 +21,7 @@ spec:
     fsGroup: 1000
 
   nodePools:
-  - name: master
+  - name: MasterPool
     replicas: 3
 
     roles:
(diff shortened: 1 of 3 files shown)


docs/quick-start/es-cluster-demo.yaml
```diff
@@ -21,7 +21,7 @@ spec:
     fsGroup: 1000
 
   nodePools:
-  - name: master
+  - name: MasterPool
```
munnerz: Don't think this was meant to be in this PR

hack/e2e.sh
```diff
@@ -109,6 +109,17 @@ function test_elasticsearchcluster() {
...
+
+function test_logs() {
+    if kubectl logs deployments/nav-e2e-navigator-controller \
+            | grep '^E' \
+            | grep -v 'reflector.go:205'; then
+        fail_test "Unexpected errors in controller logs"
+    fi
+}
+
+test_logs
```
munnerz: As much as this does make it easier to read logs, I think it could also obscure information and make debugging tests harder, especially if the developer isn't aware that we filter logs out. I think the kubernetes `gubernator` (for displaying logs from builds) can handle this sort of thing fairly well, although not 100% sure.

I'd prefer to improve the way we view logs, instead of altering the logs we're producing if that makes sense.

wallrj:
Closed in favour of https://github.com/jetstack/navigator/pull/100

Issue #95, opened 2017-11-02
RBAC prevents patching of event counts


wallrj:
When an event occurs repeatedly, the count is incremented by patching the event configuration.
The RBAC policy in our helm chart prevents this.

```
[...]

PR #100, opened 2017-11-03
Make e2e tests fail when controller logs contain unexpected errors

wallrj:
Filter the logs for errors that are not expected.
e.g. Errors caused by the controller attempting to list navigator API resources before the navigator API has been registered.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -109,6 +109,24 @@ function test_elasticsearchcluster() {
 
 test_elasticsearchcluster
 
+function ignore_expected_errors() {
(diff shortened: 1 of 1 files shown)


wallrj:
In https://github.com/jetstack/navigator/pull/96#discussion_r148523729 @munnerz wrote:
> As much as this does make it easier to read logs, I think it could also obscure information and make debugging tests harder, especially if the developer isn't aware that we filter logs out.

I still print the entire logs if the tests fail.
This new test is intended to alert us to unexpected errors in the log files, such as those that were caused by the missing `event: patch` verb in #95

PR #111, opened 2017-11-09
Allow the log level to be configured in the helm chart

wallrj:
* Adds a logLevel parameter for the apiserver and the controller deployments.
* Sets the logLevel to 100 by default
* Additional e2e test for unexpected apiserver errors.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/contrib/charts/navigator/values.yaml
+++ b/contrib/charts/navigator/values.yaml
@@ -17,7 +17,7 @@ apiserver:
     repository: jetstackexperimental/navigator-apiserver
     tag: latest
     pullPolicy: Always
-
+  logLevel: 100
 controller:
   ## Optional: namespace to watch for resources in. This can be used when RBAC
   ## restricts you to a single namespace.
@@ -28,7 +28,7 @@ controller:
     repository: jetstackexperimental/navigator-controller
     tag: latest
     pullPolicy: Always
-
+  logLevel: 100
 resources:
   requests:
     cpu: 50m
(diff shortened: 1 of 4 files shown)


contrib/charts/navigator/values.yaml
```diff
@@ -28,7 +28,7 @@ controller:
     repository: jetstackexperimental/navigator-controller
     tag: latest
     pullPolicy: Always
-
+  logLevel: 100
```
munnerz: 100 is *super* high. We'll get binary logs of every HTTP request with this. `4` is generally navigators own 'debug' level, 6 includes a log of HTTP requests (and their headers iirc) from all the underlying k8s clientsets etc.

Let's keep the default here as the actual default, `2` (or just omit it entirely if not specified in values) and then we can override it to 100 during our e2e tests (like you already are in this PR 😄)

munnerz: (although I still think 100 is *super* high for e2e tests even)

munnerz:
Just in response to your commit message ("Force highest log level in e2e tests") - there is actually no limit to the level (bar maybe the maximum int limit). There *could* be some messages logged above 100, although I'm not sure 😄

contrib/charts/navigator/values.yaml
wallrj: Ok. I couldn't find any documentation about the log levels so I chose 100 since that's what the controller is currently hard coded to.

PR #113, opened 2017-11-09
Refactoring genericpilot implementation

munnerz:

Improvements to genericpilot package to encompass full process lifecycle

Still WIP

**Release note**:
```release-note
NONE
```

/assign

### DIFF
--- a/pkg/apis/navigator/types.go
+++ b/pkg/apis/navigator/types.go
@@ -113,22 +113,24 @@ type PilotList struct {
 }
 
 type PilotSpec struct {
-	Phase         PilotPhase
 	Elasticsearch *PilotElasticsearchSpec
 }
 
 type PilotPhase string
 
 const (
-	PilotPhaseStarted PilotPhase = "Started"
-	PilotPhaseStopped PilotPhase = "Stopped"
+	PilotPhasePreStart  PilotPhase = "PreStart"
+	PilotPhasePostStart PilotPhase = "PostStart"
+	PilotPhasePreStop   PilotPhase = "PreStop"
+	PilotPhasePostStop  PilotPhase = "PostStop"
 )
 
 type PilotElasticsearchSpec struct {
 }
 
 type PilotStatus struct {
-	Conditions []PilotCondition
+	LastCompletedPhase PilotPhase
+	Conditions         []PilotCondition
 }
 
 // PilotCondition contains condition information for a Pilot.
--- a/pkg/controllers/elasticsearch/role/resources.go
+++ b/pkg/controllers/elasticsearch/role/resources.go
@@ -25,8 +25,18 @@ func roleForCluster(c *v1alpha1.ElasticsearchCluster) *rbacv1beta1.Role {
 			},
 			{
 				APIGroups: []string{navigator.GroupName},
-				Verbs:     []string{"update", "get"},
-				Resources: []string{"pilots", "elasticsearchclusters"},
+				Verbs:     []string{"*"},
+				Resources: []string{
+					"pilots",
+					"pilots/status",
+				},
+			},
+			{
+				APIGroups: []string{navigator.GroupName},
+				Verbs:     []string{"get", "list", "watch"},
+				Resources: []string{
+					"elasticsearchclusters",
+				},
 			},
 		},
 	}
--- a/pkg/controllers/elasticsearch/role/resources.go
+++ b/pkg/controllers/elasticsearch/role/resources.go
@@ -25,17 +25,17 @@ func roleForCluster(c *v1alpha1.ElasticsearchCluster) *rbacv1beta1.Role {
 			},
 			{
 				APIGroups: []string{navigator.GroupName},
-				Verbs:     []string{"*"},
+				Verbs:     []string{"get", "list", "watch"},
 				Resources: []string{
 					"pilots",
-					"pilots/status",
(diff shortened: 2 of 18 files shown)


pkg/pilot/elasticsearch/v5/options.go
```diff
@@ -144,8 +144,15 @@ func (o *PilotOptions) Run(stopCh <-chan struct{}) error {
...
-	go o.sharedInformerFactory.Start(stopCh)
+	// set the genericPilot on
+	o.pilot.genericPilot = genericPilot
+
+	// create a new stopCh just for the factory so the factory continues to
+	// receive updates after the process has been signaled to exit
+	stopInformers := make(chan struct{})
+	defer close(stopInformers)
+	// start the shared informer factory
+	go o.sharedInformerFactory.Start(stopInformers)
```
wallrj: ❓ So this is the case where kubelet has sent sigterm to the pilot, but we need to continue running while we evict shards from this database node etc and update the Pilot status.
Perhaps expand a little on the comment above.

pkg/pilot/genericpilot/controller.go
```diff
@@ -38,19 +40,23 @@ func (e *GenericPilot) processNextWorkItem() bool {
...
 }
 
-func (g *GenericPilot) sync(key string) (err error) {
+func (g *GenericPilot) sync(obj interface{}) (err error) {
+	var key string
+	var ok bool
+	if key, ok = obj.(string); !ok {
+		glog.Errorf("Unexpected non-string item passed to sync: %#v", obj)
+		return nil
+	}
```
wallrj: ❓ Seems like the signature should remain `sync(key string)` and perform this check in the caller....and just blow up, since this should always be `string` as far as I can see.

Thats how Kubernetes Deployment controller deals with it: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/deployment_controller.go#L472

pkg/pilot/genericpilot/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,70 @@
...
+						t.Errorf("expected obj '%+v' but got obj '%+v'", test.obj, obj)
+					}
+				})
+				queue.Add(test.obj, test.duration)
+			}
+		}(test))
+	}
+
+	wg.Wait()
+}
```
wallrj: These tests could be made faster if you injected a `clock` object into the queue rather than waiting for time to elapse.

wallrj: https://github.com/kubernetes/apimachinery/tree/master/pkg/util/clock

pkg/pilot/genericpilot/util.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/golang/glog"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
+)
+
+// IsThisPilot will return true if 'pilot' corresponds to the Pilot resource
+// for this pilot.
+func (g *GenericPilot) IsThisPilot(pilot *v1alpha1.Pilot) bool {
+	return pilot.Name == g.Options.PilotName && pilot.Namespace == g.Options.PilotNamespace
```
wallrj: ❓ Should we check for same owner here too?

munnerz: The Pilot is not aware of the owner here, as we don't pass 'owning' information through. Instead, we pass through a pilot 'name and namespace' as args, and have the Pilots perform this comparison to 'find itself'.

After finding itself, it *then* checks the `ownerReferences` field to find out who the owning resource is.

pkg/controllers/elasticsearch/role/resources.go
```diff
@@ -25,8 +25,18 @@ func roleForCluster(c *v1alpha1.ElasticsearchCluster) *rbacv1beta1.Role {
 			},
 			{
 				APIGroups: []string{navigator.GroupName},
-				Verbs:     []string{"update", "get"},
-				Resources: []string{"pilots", "elasticsearchclusters"},
+				Verbs:     []string{"*"},
+				Resources: []string{
+					"pilots",
```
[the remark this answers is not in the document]
munnerz: 👍

pkg/apis/navigator/types.go
```diff
@@ -113,22 +113,24 @@ type PilotList struct {
...
 
 type PilotPhase string
 
 const (
-	PilotPhaseStarted PilotPhase = "Started"
-	PilotPhaseStopped PilotPhase = "Stopped"
+	PilotPhasePreStart  PilotPhase = "PreStart"
+	PilotPhasePostStart PilotPhase = "PostStart"
+	PilotPhasePreStop   PilotPhase = "PreStop"
+	PilotPhasePostStop  PilotPhase = "PostStop"
```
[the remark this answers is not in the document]
munnerz: Done

pkg/pilot/genericpilot/scheduler/scheduler_test.go
munnerz: 👍 agreed - this will however take some significant changes to the scheduler package (as discussed) so may defer this in the interest of time, so we can work on Pilot implementations.

Issue #110, opened 2017-11-08
RBAC roles and bindings installed by ./hack/prepare-e2e.sh masks errors and missing configuration in the helm RBAC  template


wallrj:
Can we remove any of that stuff from prepare-e2e.sh and move it into the chart?

And I think we need to figure out how to force kubectl to connect to the apiserver as an authenticated user.

/kind bug

munnerz:
I think we could probably move this into the chart:

```yaml
# Create a ClusterRole to work with ElasticsearchCluster resources
- apiVersion: rbac.authorization.k8s.io/v1beta1
  kind: ClusterRole
  metadata:
    name: navigator:authenticated
  # this rule defined on the role for specifically the
  # namespace-lifecycle admission-controller
  rules:
  - apiGroups: ["navigator.jetstack.io"]
    resources: ["elasticsearchclusters", "pilots"]
    verbs:     ["get", "list", "watch", "create", "update", "delete"]
- apiVersion: rbac.authorization.k8s.io/v1beta1
  kind: ClusterRoleBinding
  metadata:
    name: "navigator:authenticated"
  roleRef:
    apiGroup: rbac.authorization.k8s.io
    kind: ClusterRole
    name: navigator:authenticated
  subjects:
  - kind: Group
    name: system:authenticated
    apiGroup: rbac.authorization.k8s.io
  - kind: Group
    name: system:unauthenticated
    apiGroup: rbac.authorization.k8s.io
```

but not the other bits.

afaik kubectl *is* connecting as an authenticated user - I think the error message just masks what the actual username is or something. Otherwise more requests than just the ones we see failing would be failing (plus also, minikube configures kubectl to talk to the apiserver on port 8443, which is a secure port which requires auth)

munnerz:
Closed by #122

PR #115, opened 2017-11-09
Remove unnecessary apiserver flags

wallrj:
These were somehow preventing the navigator API server from seeing the user that
made the original request to the k8s API server.

Fixes: #110 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,14 +20,6 @@ jobs:
       - make BUILD_TAG=latest build e2e-test
 
     - stage: test
-      env:
-      - KUBERNETES_VERSION=v1.7.0
-      before_script:
-      - ./hack/install-e2e-dependencies.sh
-      script:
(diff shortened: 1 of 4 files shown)


wallrj:
@munnerz  See https://github.com/jetstack/navigator/pull/116 for an example of how Kubernetes 1.7 also has the same RBAC issues which were being masked by the prepare-e2e.sh RBAC configuration.

Hopefully #109 is resolved in 1.8.{1,2,3}  so I suggest we merge this. And upgrade our tests to use Kubernetes 1.8.3 as soon as it's supported by Minikube: https://github.com/kubernetes/minikube/issues/2156

wallrj:
@munnerz Also see https://github.com/jetstack/navigator/pull/120 where I retry minikube start if the API server fails to respond within 60s.

Issue #127, opened 2017-11-14
Cassandra nodes can't connect to seed provider


wallrj:
```
WARN  11:58:05 Seed provider couldn't lookup host cass-cassandra-1510658740-31487-cassandra-ringnodes-0.cass-cassandra-1510658740-31487-cassandra.test-cassandracluster-1510658740-31487.svc.cluster.local
The seed provider lists no seeds.
Exception (org.apache.cassandra.exceptions.ConfigurationException) encountered during startup: The seed provider lists no seeds.
```

1. The service we're creating needs to be headless.
2. Or we could use KubernetesSeedProvider from: https://github.com/kubernetes/examples/tree/master/cassandra/java
3. Or we look at the other cloud specific "snitches" which give cassandra more information about the datacentre, zone and region of cassandra nodes: http://docs.datastax.com/en/archived/cassandra/3.x/cassandra/architecture/archSnitchesAbout.html 

Probably do 1 first and then follow up with 2 or 3

/kind bug

wallrj:
Fixed in https://github.com/jetstack/navigator/pull/128

PR #128, opened 2017-11-14
Create a headless service so that other nodes can lookup the seed provider 

wallrj:
Fixes: #127 

**Release note**:
```release-note
NONE
```


wallrj:
Logs show:
```
I1116 12:09:56.387] INFO  12:09:48 Starting Messaging Service on /172.17.0.8:7000 (eth0)
I1116 12:09:56.387] WARN  12:09:48 No host ID found, created 471089dd-269d-498f-a96d-1c8cb577b79c (Note: This should happen exactly once per node).
I1116 12:09:56.387] INFO  12:09:49 Handshaking version with cass-cassandra-1510834028-24677-cassandra-ringnodes-0.cass-cassandra-1510834028-24677-cassandra-seedprovider.test-cassandracluster-1510834028-24677.svc.cluster.local/172.17.0.7
I1116 12:09:56.387] WARN  12:09:54 Seed gossip version is -2147483648; will not connect with that version
I1116 12:09:56.387] INFO  12:09:54 Cannot handshake version with cass-cassandra-1510834028-24677-cassandra-ringnodes-0.cass-cassandra-1510834028-24677-cassandra-seedprovider.test-cassandracluster-1510834028-24677.svc.cluster.local/172.17.0.7
I1116 12:09:56.388] INFO  12:09:54 Handshaking version with cass-cassandra-1510834028-24677-cassandra-ringnodes-0.cass-cassandra-1510834028-24677-cassandra-seedprovider.test-cassandracluster-1510834028-24677.svc.cluster.local/172.17.0.7

```

Perhaps I just need to give the second node longer to connect to the first.
This test passes on my GKE cluster.

pkg/controllers/cassandra/service/testing/testing.go
```diff
@@ -0,0 +1,59 @@
...
+type FixtureFactory func(t *testing.T) *casstesting.Fixture
+type ServiceFactory func(f *casstesting.Fixture) *apiv1.Service
+
+func RunStandardServiceTests(t *testing.T, newFixture FixtureFactory, newService ServiceFactory) {
+	t.Run(
+		"service created",
+		func(t *testing.T) {
+			f := newFixture(t)
+			f.Run()
+			f.AssertServicesLength(1)
```
munnerz: Can we perform any checks to ensure the created service has the required ports/type etc. set? Or is this done elsewhere? 😄

pkg/controllers/cassandra/service/util/util.go
```diff
@@ -0,0 +1,60 @@
...
+type ServiceUpdater func(cluster *v1alpha1.CassandraCluster, service *apiv1.Service) *apiv1.Service
+
+func SyncService(
+	cluster *v1alpha1.CassandraCluster,
+	kubeClient kubernetes.Interface,
+	serviceLister corelisters.ServiceLister,
+	createService ServiceCreator,
+	updateService ServiceUpdater,
+) error {
+	svc := createService(cluster)
```
munnerz: Do we need a creator as well as an updater? In my mind, calling `updateService` with an `&apiv1.Service{}` should have the same effect as calling `createService`?

munnerz: Ah - looks like that's what is always happening, but then we pass a Creator/Updater func to this function. Is there any reason for this? Do we behave differently ever? It'd be ideal to simplify it if we can.

pkg/controllers/cassandra/util/util.go
```diff
@@ -35,6 +36,14 @@ func NodePoolResourceName(c *v1alpha1.CassandraCluster, np *v1alpha1.CassandraCl
...
 }
 
+func SeedProviderServiceName(c *v1alpha1.CassandraCluster) string {
+	return fmt.Sprintf("%s-seedprovider", ResourceBaseName(c))
+}
+
+func CqlServiceName(c *v1alpha1.CassandraCluster) string {
+	return fmt.Sprintf("%s-cql", ResourceBaseName(c))
+}
+
```
munnerz: Note (but not required to be fixed in this PR, as I think it's a wider problem) - I've run into problems in cert-manager where resource names become too long as we just keep appending to things. We should try and work out a 'grand unified way' to generate resource names predictably.

hack/e2e.sh
```diff
@@ -241,16 +250,8 @@ function test_logged_errors() {
...
-if [[ "${FAILURE_COUNT}" -gt 0 ]]; then
-    kubectl get po -o yaml
-    kubectl describe po
-    kubectl get svc -o yaml
-    kubectl describe svc
-    kubectl get apiservice -o yaml
-    kubectl describe apiservice
-    kubectl logs -c apiserver -l app=navigator,component=apiserver
-    kubectl logs -c controller -l app=navigator,component=controller
-    kubectl logs -c etcd -l app=navigator,component=apiserver
```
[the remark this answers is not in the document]
wallrj: I added `kubectl cluster-info dump --all-namespaces` which appears to provide all this information and more. For example, it gives me the logs of all the cassandra pods which is how I was able to see that the test wasn't allowing long enough for the second node to join the cluster.

pkg/controllers/cassandra/service/testing/testing.go
wallrj: There are e2e checks for the service ports.

pkg/controllers/cassandra/service/util/util.go
wallrj: I need the create function for generating services for the unit tests. But you're right, it can probably be simplified in a future branch.

munnerz:
Could you squash this into 1 commit before merge? 😄

munnerz:
(then all good! I don't want to say the magic phrase for fear this will auto merge... 🙄 )

wallrj:
> Could you squash this into 1 commit before merge?

@munnerz Done. Thanks for the review.

PR #129, opened 2017-11-14
Initial Cassandra support

wallrj:
**Release note**:
```release-note
NONE
```


munnerz:
As well as rebasing, we really should try and tidy up some of these commit messages by squashing. This may be harder to do now, but we could start by coming up with a way to retrospectively squash PRs perhaps.

munnerz:
Approving so we don't have this divergence of branches. Work still needs to be done to support cassandra, but this should make development easier at this early stage.

/lgtm
/approve

munnerz:
@wallrj seems there's a few issues here that intermittently occur. Are you able to look into these? I know one of those is the Elasticsearch role issue, which is something known, but the latest test run seemed to fail on some cassandra related things too 🤔

wallrj:
@munnerz https://github.com/jetstack/navigator/pull/155 is needed to fix flaky readiness probe tests.
Approve that and then this should pass more consistently.

hack/e2e.sh
```diff
@@ -141,13 +141,122 @@ function test_elasticsearchcluster() {
...
+    # No queries are performed.
+    # stdin=false (the default) ensures that cqlsh does not go into interactive
+    # mode.
+    # XXX: This uses the standard Cassandra Docker image rather than the
+    # gcr.io/google-samples/cassandra image used in the Cassandra chart, becasue
+    # cqlsh is missing some dependencies in that image.
+    kubectl \
+        run \
+        "cql-responding-${RANDOM}" \
+        --namespace="${USER_NAMESPACE}" \
```
munnerz: USER_NAMESPACE here should be ${namespace}

wallrj: Ah, yeah, that's a mistake, but it doesn't affect the test results.
`USER_NAMESPACE` is redefined as a local in `test_cassandracluster`  and it'll be using that.
I propose I fix this in a followup branch rather than have to go through another cycle of E2E tests.

PR #131, opened 2017-11-14
Set Elasticsearch document count on Pilot resource

munnerz:

Sets the `pilot.status.elasticsearch.documents` field periodically

This needs some e2e tests. I'm going to attempt to modify the existing tests to boot a single node Elasticsearch cluster.

**Release note**:
```release-note
Set the Elasticsearch node document count on the Pilot resource
```


munnerz:
This now depends on #134 being merged

munnerz:
@wallrj this is ready for review :)

wallrj:
@munnerz Any chance you could merge in master, now that the namespaced sharedinformer branch is merged?

munnerz:
Ah - did not realise I needed to!
Will do when I am at my computer!

munnerz:
@wallrj rebased

hack/install-e2e-dependencies.sh
```diff
@@ -13,6 +13,7 @@ minikube start \
      --vm-driver=kvm \
      --kubernetes-version="$KUBERNETES_VERSION" \
      --bootstrapper=kubeadm \
+     --memory 4096 \
```
munnerz: @wallrj this should hopefully solve some performance issues you've had with cassandra running in e2e tests

pkg/pilot/elasticsearch/v5/pilot.go
```diff
@@ -47,7 +47,6 @@ func NewPilot(opts *PilotOptions) (*Pilot, error) {
 		pilotInformerSynced:     pilotInformer.Informer().HasSynced,
 		esClusterLister:         esClusterInformer.Lister(),
 		esClusterInformerSynced: esClusterInformer.Informer().HasSynced,
-		localESClient:           &elastic.Client{},
```
munnerz: This was causing ```statsList, err := p.localESClient.NodesStats().NodeId(elasticsearchLocalNodeID).Do(context.Background())``` to panic, as the `localESClient.client` was nil as it wasn't constructed properly.

Instead, we now check to see if the `localESClient` is nil: https://github.com/jetstack/navigator/pull/131/files#diff-524ec275431e2337616e927829c48547R26

This has to be done because here: https://github.com/munnerz/navigator/blob/0fa9f757d2280bdf626653cbf07b610c4d37d9ae/pkg/pilot/elasticsearch/v5/pilot.go#L53

we continuously attempt to obtain an elastic client. Within the `NewClient` constructor, elastic.Client will attempt to connect to Elasticsearch and verify it's okay to do so, and returns an error if it's not, hence this 'retrying' behaviour.

hack/e2e.sh
```diff
@@ -102,7 +102,7 @@ function test_elasticsearchcluster() {
     # Create and delete an ElasticSearchCluster
     if ! kubectl create \
             --namespace "${USER_NAMESPACE}" \
-            --filename "${ROOT_DIR}/docs/quick-start/es-cluster-demo.yaml"; then
+            --filename "${SCRIPT_DIR}/testdata/es-cluster-test.yaml"; then
```
wallrj: ❓ I'm a bit sad that we'll no longer be testing the documented resource. The documentation will now diverge from the test data and maybe become invalid.

hack/install-e2e-dependencies.sh
wallrj: Great. Although my builds have been quite reliable lately.

pkg/pilot/elasticsearch/v5/controller.go
```diff
@@ -1,11 +1,41 @@
...
+	if pilot.Status.Elasticsearch == nil {
+		pilot.Status.Elasticsearch = &v1alpha1.ElasticsearchPilotStatus{}
+	}
+	// set the ES document count to nil to indicate an unknown number of
+	// documents in case obtaining the document count fails. This prevents the
+	// node being shut down if it is not safe to do so.
+	pilot.Status.Elasticsearch.Documents = nil
+	if p.localESClient == nil {
+		return fmt.Errorf("local elasticsearch client not available")
+	}
```
wallrj: Thanks for explaining this, in https://github.com/jetstack/navigator/pull/131#pullrequestreview-77386340.
Did you see https://github.com/olivere/elastic/wiki/Retrier-and-Backoff? Maybe if you configure that, the `NewClient` won't immediately fail.
Also consider creating a new `NewSimpleClient` here each time: https://github.com/olivere/elastic/blob/release-branch.v6/client.go#L331

pkg/pilot/elasticsearch/v5/controller.go
```diff
@@ -1,11 +1,41 @@
...
+	if !p.genericPilot.IsThisPilot(pilot) {
+		return nil
+	}
+	if pilot.Status.Elasticsearch == nil {
+		pilot.Status.Elasticsearch = &v1alpha1.ElasticsearchPilotStatus{}
+	}
+	// set the ES document count to nil to indicate an unknown number of
+	// documents in case obtaining the document count fails. This prevents the
+	// node being shut down if it is not safe to do so.
+	pilot.Status.Elasticsearch.Documents = nil
```
wallrj: So the supplied `Pilot` gets modified, even if this function errs?

pkg/pilot/elasticsearch/v5/controller.go
wallrj: `p.localESClient` is also called in `probes.go` . Does that need the same nil pointer check?

hack/e2e.sh
munnerz: Yep it will - if we want to test docs though, I think we should have separate tests. Our test fixtures are going to necessarily diverge from docs, due to disabling persistence, creating single node clusters, etc.

Our tests should be 'narrow' and test individual parts of Navigator's behaviour, so we can end out with an ElasticsearchCluster manifest that is specific to testing say, persistence. Another that's specific to testing a custom port is set, and so on. If all of these checks pass, then we know that the documented examples also pass (because we cover the full set of functionality described in the manifest).

pkg/pilot/elasticsearch/v5/controller.go
munnerz: Yes

pkg/pilot/elasticsearch/v5/controller.go
munnerz: Yep it will also need that check - thanks for pointing it out. I'd like to come up with a nicer way to do this :(

& thanks for the links - I'll have a read of those!

munnerz:
> Looks good. I think you've explained that you intend to examine this new Documents count from the ElasticSearch node poole controller, to know when it's safe to scale down the statefulset.
> Wouldn't a GenericPilot.Decomissioned attribute be enough? Perhaps we already discussed that during the design discussion.

Yes it would be enough, however it's up to the 'leader pilot' to set this field on the node after evaluating the 'state of the world'. Because setting the `decommissioned` field to true will result in this Pilot being shut down (soon). The document count right now might be the only heuristic we check, but in future we may also check the document count, as well as say, the number of nodes currently ready. This allows the ES Pilot 'leader' to prevent a scale down occurring until it deems it is safe to do so.

Issue #136, opened 2017-11-16
Cassandra CQL service should only route to nodes when they're ready


wallrj:
/kind bug

wallrj:
Merged.

PR #137, opened 2017-11-16
ReadinessProbes and LivenessProbes

wallrj:
ReadinessProbes and LivenessProbes

* Added ReadinessProbe that calls /ready-probe.sh
* Added a LivenessProbe that also calls /ready-probe.sh, but with a delay, to allow the database to initialise and connect to its peers.
* Added an E2E test for readiness probe, that stops the processes inside the first cassandra pod and tests that the CQL service is still reachable, i.e. the readiness probe has failed and the service is no longer routing to that pod.
* Use cqlsh to better test that the cassandra CQL port is responding. Netcat only tested that a TCP connection could be established, which also works when the listening process has been stopped.
* Set the `PublishNotReadyAddresses` field on the headless service so that node hostnames names get published regardless of the ReadinessProbe. This allows joining nodes to find peers regardless of whether they are ready.
* Also set the "tolerate-unready-endpoints" annotation for backwards compatibility with Kubernetes <= 1.7

Fixes: #136

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cassandra/service/seedprovider/resource.go
+++ b/pkg/controllers/cassandra/service/seedprovider/resource.go
@@ -7,6 +7,10 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 )
 
+const (
+	TolerateUnreadyEndpointsAnnotationKey = "service.alpha.kubernetes.io/tolerate-unready-endpoints"
+)
+
 func ServiceForCluster(
 	cluster *v1alpha1.CassandraCluster,
 ) *apiv1.Service {
@@ -30,5 +34,19 @@ func updateServiceForCluster(
 			Port: 65535,
 		},
 	}
+	// This ensures that DNS names are published regardless of whether the
+	// Cassandra pod ReadinessProbes (listening on their CQL port).
+	// It won't handle CQL connections until it has successfully connected and
+	// negotiated with a seed host
+	service.Spec.PublishNotReadyAddresses = true
+	// XXX: This annotation is only necessary for Kubernetes <=1.7, which do not
+	// pay attention to the field above.
+	// Remove it when we no longer support those versions.
+	annotations := service.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[TolerateUnreadyEndpointsAnnotationKey] = "true"
+	service.SetAnnotations(annotations)
 	return service
 }
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -141,19 +141,29 @@ function test_elasticsearchcluster() {
 
 test_elasticsearchcluster
 
-function kube_service_responding() {
+function cql_connect() {
     local namespace="${1}"
     local host="${2}"
     local port="${3}"
-    kubectl run \
-            --namespace="${namespace}" \
-            "kube-service-responding-${RANDOM}" \
-            --stdin=true\
-            --rm \
-            --restart=Never \
-            --image alpine \
-            -- \
-            nc -w 5 -v "${host}" "${port}" <<< ping
+    # Attempt to negotiate a CQL connection.
+    # No queries are performed.
+    # stdin=false (the default) ensures that cqlsh does not go into interactive
+    # mode.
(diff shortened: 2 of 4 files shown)


pkg/controllers/cassandra/service/seedprovider/resource.go
```diff
@@ -0,0 +1,39 @@
...
+	service.Spec.Ports = []apiv1.ServicePort{
+		{
+			Port: 65535,
+		},
+	}
+	// This ensures that DNS names are published regardless of whether the
+	// Cassandra pod ReadinessProbes (listening on their CQL port).
+	// It won't handle CQL connections until it has successfully connected and
+	// negotiated with a seed host
+	service.Spec.PublishNotReadyAddresses = true
```
wallrj: See also https://github.com/kubernetes/examples/issues/89

munnerz: This field has only been promoted in 1.8+ - we should also set the `service.alpha.kubernetes.io/tolerate-unready-endpoints ` annotation, which works back to 1.5 or 1.6 or something :smile:

Awesome to know that it's been promoted to a field now though!

wallrj: Isn't that handled by API version conversion? Otherwise how can they add a field to an API without changing the version?

munnerz: Yep it is - but if the target API server is not aware of the field, it will silently drop it. So if you submit the manifest produced above to a 1.7 APIServer, because 1.7 has never contained this field, it will be silently dropped.

If on the other hand you submit this to a 1.8 API server, it will automatically 'convert' it by additionally setting the aforementioned annotation. But this requires that you submit it to a 1.8 API server. Thus, we should set the annotation too if we want to support older k8s clusters. We should set both the field and annotation so that when 1.10 comes out and drops support for the annotation, we already support 1.10 as we set the field :smile:

wallrj: Got it. Thanks @munnerz 
I hadn't thought it through and maybe I was assuming some sort of conversion on the client side too.

munnerz:
Quick side note: can you tidy up commits a bit before merging? I guess it'll make it easier to do in chunks rather than having to tidy them all up before the main PR is merged 😄

wallrj:
> Quick side note: can you tidy up commits a bit before merging? I guess it'll make it easier to do in chunks rather than having to tidy them all up before the main PR is merged

I'm not that thrilled about squashing and rebasing my commits. I have enough trouble with git as it is.
Can't we just squash the commits when they get merged?
Or squash all the PRs to master in the order that they've been squashed to the feature branch.

munnerz:
Sure thing. But you'll need to squash it yourself as the submit queue won't do it for us.

So long as you are up to date with the branch you are merging into (here, `23-cassandra`), you should be able to:
```
git rebase -i 23-cassandra
```

Then change 'pick' to 'squash' on every commit you want to squash together. Those commits will be squash into the commit *above* them, so you'll probably leave the commit at the top as `pick`. It should look something like this:

```
pick a0af375 Update Kubernetes dependency pinning
s 67d172e dep ensure -update
s eb658e2 Update to use new generate-groups.sh script
s 00b833c Run ./hack/update-codegen.sh
s 196450e Clean up old api related files
s e9d3afc Update imports for new client/informers
s 7526a1a Update Makefile for new codegen
s daabedc Update .gitignore
s e279350 Add verify-codegen script

# Rebase 59148fe..e279350 onto 59148fe (9 commands)
#
# Commands:
...
```

(note: `s` is an alias for `squash`, and this is also an example done on another repo).

In this case, all of the commits with 's' before them would be squashed into "Update Kubernetes dependency pinning". You will also be given the opportunity to write a new commit message before the rebase is finished.

Let me know if you run into any issues! 😄

pkg/controllers/cassandra/service/seedprovider/resource.go
wallrj: Done.

Issue #141, opened 2017-11-22
Create a Cassandra Pilot


wallrj:
/kind feature

See #140

PR #142, opened 2017-11-22
Cassandra pilot

wallrj:
* A pilot command for managing a Cassandra subprocess
* Installed into the Cassandra container via an InitContainer
* Watches for a Pilot resource with the name and namespace, which are supplied as command line arguments.
* Then launches the Cassandra subprocess.

Fixes: #141 

**Release note**:
```release-note
NONE
```


pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -52,6 +52,48 @@ func StatefulSetForCluster(
...
+										Command: []string{
+											"/usr/bin/timeout",
+											"5",
+											"/ready-probe.sh",
+										},
+									},
+								},
+								InitialDelaySeconds: 60,
+								// XXX Kubernetes ignores the TimeoutSeconds for Exec probes.
+								// See https://github.com/kubernetes/kubernetes/issues/26895
```
munnerz: Just FYI - as GenericPilot exposes HTTP health checks, we should be able to work around this once this becomes a Pilot probe 😄

hack/e2e.sh
```diff
@@ -27,6 +27,10 @@ mkdir -p $TEST_DIR
 
 source "${SCRIPT_DIR}/libe2e.sh"
 
+# Mandatory environment variables
+: ${CHART_VALUES:?}
+: ${CHART_VALUES_CASSANDRA:?}
```
munnerz: Given we can now toggle which e2e tests to run (i.e. I could not run cassandra e2e), this var isn't *always* required. I can see it's defaulted in the Makefile itself, but would it make sense to move some of these defaults into this file? (perhaps including CHART_VALUES)?

hack/e2e.sh
```diff
@@ -37,10 +41,6 @@ function debug_navigator_start() {
 }
 
 function helm_install() {
-    if [ "${CHART_VALUES}" == "" ]; then
-        echo "CHART_VALUES must be set";
-        exit 1
-    fi
```
munnerz: Is this no longer required?

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -123,6 +149,13 @@ func StatefulSetForCluster(
 									ContainerPort: util.DefaultCqlPort,
 								},
 							},
+							VolumeMounts: []apiv1.VolumeMount{
+								{
+									Name:      sharedVolumeName,
+									MountPath: sharedVolumeMountPath,
+									ReadOnly:  false,
```
munnerz: (nit) can we make this ReadOnly: true? Although ES also has readOnly: false set, so can be a follow up/pending discussion

wallrj: Good idea. Done. For ElasticSearch too.

hack/e2e.sh
wallrj: Nope. The `: ${CHART_VALUES:?}` causes the script to exit with almost that exact message if the variable isn't set.

hack/e2e.sh
wallrj: Good idea. Done.

PR #144, opened 2017-11-22
Rely on externally provisioned cluster for e2e tests

munnerz:
**What this PR does / why we need it**:

Switches us to take advantage of https://github.com/jetstack/test-infra/pull/79 - where the minikube vm is provisioned by some external source.

**Release note**:
```release-note
NONE
```

/assign

PR #146, opened 2017-11-23
Delete Elasticsearch test namespace after e2e run

munnerz:

This PR cleans up the namespace used during tests after the test has run. Previously, this would not happen due errors when genericpilot attempts to exit after it's Pilot resource has been deleted.

In order to work around this, I have added a new cachedThisPilot field to GenericPilot, which is used in favour of the 'up to date' Pilot resource from the API server. This allows GenericPilot to continue operating in the event of it's own Pilot being deleted, or in the case of some other issue that may arise.

Moving forward, we need to ensure that *all* our e2e tests clean up unused resources upon completion, as we cannot infinitely increase the size of our build VMs just so we can keep logs hanging around. If we need logs from pods, we should retrieve them as part of the tear-down for that particular test suite.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -134,6 +134,9 @@ function test_elasticsearchcluster() {
             --all; then
         fail_test "Failed to delete elasticsearchcluster"
     fi
+    if ! kube_delete_namespace_and_wait "${USER_NAMESPACE}"; then
+        fail_test "Failed to delete test namespace"
+    fi
 }
 
 test_elasticsearchcluster
--- a/pkg/pilot/genericpilot/controller.go
+++ b/pkg/pilot/genericpilot/controller.go
@@ -72,7 +72,13 @@ func (g *GenericPilot) sync(key string) (err error) {
 	pilot, err := g.pilotLister.Pilots(namespace).Get(name)
 	if apierrors.IsNotFound(err) {
 		glog.Infof("Pilot %q has been deleted", key)
-		return nil
+		if !g.isThisPilot(name, namespace) || g.cachedThisPilot == nil {
+			return nil
+		}
+		glog.Infof("Using cached pilot resource for %q", key)
+		pilot = g.cachedThisPilot
+		// set err to nil so the following block does not return err
+		err = nil
 	}
 	if err != nil {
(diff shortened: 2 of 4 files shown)


munnerz:
Regarding logs - yep I think this is a wider problem to be honest. When we switch to ginkgo, we can do something more like what cert-manager is doing, where we grab the information in a cleanup action after each suite (and before the actual resources are torn down). We can then at the end do a cluster-dump, which will return *only* cluster wide info (and no info specific to tests, as that will be handled in the respective tests).

For now however, cluster-info dump *will* include ES logs *provided* it was ES that failed (as the cleanup won't have run), which seems sufficient for the time being.

Those errors look like RBAC in kube-system isn't set up properly 🤔 . I did remove the "grant all kube-system service accounts cluster-admin permissions" hack during #144. This could have led to the above errors.

I'll update the PR description now to explain why we're doing this, and that text will be put into the merge commit. Each individual commit message explains what that one commit is doing (e.g. `Allow reconcillation of 'this pilot' in face of API errors`).

wallrj:
> For now however, cluster-info dump will include ES logs provided it was ES that failed (as the cleanup won't have run), which seems sufficient for the time being.

I don't think so. Current implementation of `fail_test` just increments a counter to signify how many of the tests failed. It deliberately doesn't stop the following tests. 

> Those errors look like RBAC in kube-system isn't set up properly 🤔 . I did remove the "grant all kube-system service accounts cluster-admin permissions" hack during #144. This could have led to the above errors.

How come you removed it? Not needed with kubeadm provisioner? If it is still important,  create a test-infra issue.

Happy to give this a lgtm if you want to address these things in followup branches.

munnerz:
@wallrj that's a shame - but there's no way we can keep the pods around long term. As our test suite grows, we could end out with 10s or hundreds of pods left around. I think grabbing the logs in an AfterSuite or the like makes more sense (just before tearing down the NS).

Regarding the RBAC hack - yeah I removed it as when I've used kubeadm to provision clusters before it's not been required, and we *shouldn't* need it. If it is causing actual issues though it obviously makes sense to include it again. This is now an issue for test-infra though as you say :)

munnerz:
& yeah I'd quite like to see this merged so we can unblock the other PRs we currently have pending. Right now basically no e2e tests pass :cry:

PR #148, opened 2017-11-23
Add genericpilot Sync unit tests

munnerz:

Adds unit tests for the genericpilot Sync method

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/pilot/genericpilot/controller_test.go
+++ b/pkg/pilot/genericpilot/controller_test.go
@@ -1,0 +1,268 @@
+package genericpilot
+
+import (
+	"reflect"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	core "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/record"
+
+	"github.com/jetstack/navigator/pkg/apis/navigator"
+	"github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
+	navfake "github.com/jetstack/navigator/pkg/client/clientset/versioned/fake"
+	informers "github.com/jetstack/navigator/pkg/client/informers/externalversions"
+	"github.com/jetstack/navigator/pkg/controllers/elasticsearch/util"
+	"github.com/jetstack/navigator/pkg/pilot/genericpilot/hook"
+	"github.com/jetstack/navigator/pkg/pilot/genericpilot/process"
+	"github.com/jetstack/navigator/pkg/pilot/genericpilot/scheduler"
+)
+
+type fixture struct {
+	// the name of the test
+	name string
+	// the key to pass to sync
+	key string
+	// whether to expect an error to be returned
+	err bool
+	// the 'thisPilot' name & namespace options for the pilot
+	thisPilotName, thisPilotNamespace string
+	// items expected to be scheduled to be re-run
+	expectedScheduledItems []interface{}
+	// actions expected to happen on the client.
+	navActions []core.Action
+	// contents of the pilotLister
+	pilotLister []*v1alpha1.Pilot
+	// list of objects to load into the clientset
+	kubeObjects []runtime.Object
+	// optional hooks resource for testing
+	hooks *hook.Hooks
+	// cachedThisPilot field on the controller. useful for testing that
+	// 'offline mode' works
+	cachedThisPilot *v1alpha1.Pilot
+	// a fake process for the controller
+	process process.Interface
+	// if set, this fixture will ensure the Pilot passed to the controllers
+	// SyncFunc deep equals this pilot
+	syncFuncExpectedPilot *v1alpha1.Pilot
+}
+
+func (f *fixture) run(t *testing.T) {
+	// initialise hooks if nil
+	if f.hooks == nil {
+		f.hooks = &hook.Hooks{}
+	}
+	// set up testing scheduled workqueue
+	expectedScheduledItems := make([]interface{}, len(f.expectedScheduledItems))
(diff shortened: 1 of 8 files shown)


munnerz:
This needs a rebase once #149 has merged

pkg/pilot/genericpilot/controller_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+func filterInformerActions(actions []core.Action) []core.Action {
+	ret := []core.Action{}
+	for _, action := range actions {
+		if len(action.GetNamespace()) == 0 &&
+			(action.Matches("list", "foos") ||
+				action.Matches("watch", "foos") ||
+				action.Matches("list", "deployments") ||
+				action.Matches("watch", "deployments")) {
+			continue
```
munnerz: This should probably be something more meaningful 🙄

munnerz: Or just remove this if block altogether as it's not useful to us right now

munnerz:
@wallrj I've refactored this a bit and added comments throughout. Now ready for review!

/test e2e

pkg/pilot/genericpilot/controller_test.go
```diff
@@ -0,0 +1,268 @@
...
+		{
+			name:                   "should not Start a subprocess when syncing a non-this pilot",
+			key:                    "namespace/not-this-pilot",
+			err:                    false,
+			thisPilotName:          "name",
+			thisPilotNamespace:     "namespace",
+			expectedScheduledItems: []interface{}{"namespace/not-this-pilot"},
+			cachedThisPilot:        newDummyPilot("name", "namespace"),
+			pilotLister:            []*v1alpha1.Pilot{newDummyPilot("not-this-pilot", "namespace")},
+			kubeObjects:            []runtime.Object{newDummyPilot("not-this-pilot", "namespace")},
```
wallrj: I see that we don't supply a process in this test, but I don't understand how this is enough to demonstrate that a subprocess is *not* started.

pkg/pilot/genericpilot/controller_test.go
```diff
@@ -0,0 +1,268 @@
...
+			name: "should reuse cached 'this' pilot resource, and should fail to update status",
+			key:  "namespace/name",
+			err:  true,
+			syncFuncExpectedPilot:  newDummyPilot("name", "namespace"),
+			thisPilotName:          "name",
+			thisPilotNamespace:     "namespace",
+			expectedScheduledItems: []interface{}{"namespace/name"},
+			cachedThisPilot:        newDummyPilot("name", "namespace"),
+			navActions:             []core.Action{newUpdatePilotStatusAction(newDummyPilot("name", "namespace"))},
+			process:                newDefaultProcessAdapter(),
```
wallrj: The test title says `should fail to update status` but the expected `navActions` include a status update.
Perhaps the test title could be reworded.

pkg/pilot/genericpilot/controller_test.go
```diff
@@ -0,0 +1,268 @@
...
+			key:  "namespace/not-this-pilot",
+			err:  false,
+			syncFuncExpectedPilot:  newDummyPilot("not-this-pilot", "namespace"),
+			thisPilotName:          "name",
+			thisPilotNamespace:     "namespace",
+			expectedScheduledItems: []interface{}{"namespace/not-this-pilot"},
+			cachedThisPilot:        newDummyPilot("name", "namespace"),
+			pilotLister:            []*v1alpha1.Pilot{newDummyPilot("not-this-pilot", "namespace")},
+			kubeObjects:            []runtime.Object{newDummyPilot("not-this-pilot", "namespace")},
+			process:                newDefaultProcessAdapter(),
```
wallrj: Again, I don't understand this test description. `should sync not-this-pilot and should not update the pilots status`.
Doesn't this show that the pilot ignores foreign / peer pilot resources.
But in that case why is the foreign / peer pilot among the `expectedScheduledItems`? 
To me, that implies that the Pilot controller has looked at the pilot, found it to be foreign and then put it back on the queue, rather than discarding it.

pkg/pilot/genericpilot/controller_test.go
```diff
@@ -0,0 +1,268 @@
...
+		},
+	}
+	// we can't use subtests yet as functions passed to our fixtures may need
+	// to reference 't' themselves
```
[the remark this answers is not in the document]
munnerz: Subtests are now used - this comment needs removing :)

pkg/pilot/genericpilot/controller_test.go
munnerz: 👍 basically, if the resource doesn't exist, it *will* attempt to update the status (hence the expected action), but will fail with an error (hence `err: true`). I could update this to say it "should attempt to update status and fail"

pkg/pilot/genericpilot/controller_test.go
munnerz: So 'not-this-pilot' in this case *is* a peer/member of the cluster, but it is not *this* pilot.

This means that it *should* be synced (right now we don't support leader election, but once we do this becomes a lot more important). But it *shouldn't* update the Status block of that Pilot as a Pilot is only responsible for updating it's *own* status block.

munnerz: It's then expected to be scheduled again so that things 'keep moving'. Perhaps in future we won't reschedule peers every 10s, but for now it saves queues getting blocked because there's been no change to any resource (so nothing happens)

pkg/pilot/genericpilot/controller_test.go
munnerz: 👍 my thinking was that the test will panic if process.{Start,Stop,Status etc..} is called, hence failing the test.

We could/should definitely extend this to actually just call `t.Fail` upon any calls to the processes methods (e.g. create a `NewFailingFakeProcess` function or something)

Issue #152, opened 2017-11-27
Create Cassandra PilotResources


wallrj:
Part of  #140

PR #153, opened 2017-11-27
Create Cassandra Pilot resources

wallrj:
* For every pod in the nodepool, create a corresponding Pilot resource.
 * Delete Pilot resources for which there is no corresponding Pod.
 * Ignore Pods that are not owned by the cluster.
 * Stop and error if we encounter Pilots with expected name, but not owned by the cluster.
    
Fixes: #152

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cassandra/pilot/pilot.go
+++ b/pkg/controllers/cassandra/pilot/pilot.go
@@ -1,0 +1,182 @@
+package pilot
+
+import (
+	"fmt"
+	"hash/fnv"
+	"reflect"
+
+	"github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
+	navigator "github.com/jetstack/navigator/pkg/client/clientset/versioned"
+	navlisters "github.com/jetstack/navigator/pkg/client/listers/navigator/v1alpha1"
+	"github.com/jetstack/navigator/pkg/controllers"
+	"github.com/jetstack/navigator/pkg/controllers/cassandra/util"
+	hashutil "github.com/jetstack/navigator/pkg/util/hash"
+	"k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	appslisters "k8s.io/client-go/listers/apps/v1beta1"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/record"
+)
+
+const (
+	HashAnnotationKey = "navigator.jetstack.io/cassandra-pilot-hash"
+)
+
+type Interface interface {
+	Sync(*v1alpha1.CassandraCluster) error
+}
+
+type pilotControl struct {
+	naviClient   navigator.Interface
+	pilots       navlisters.PilotLister
+	pods         corelisters.PodLister
+	statefulSets appslisters.StatefulSetLister
+	recorder     record.EventRecorder
+}
+
+var _ Interface = &pilotControl{}
+
+func NewControl(
+	naviClient navigator.Interface,
+	pilots navlisters.PilotLister,
+	pods corelisters.PodLister,
+	statefulSets appslisters.StatefulSetLister,
+	recorder record.EventRecorder,
+) *pilotControl {
+	return &pilotControl{
+		naviClient:   naviClient,
+		pilots:       pilots,
+		pods:         pods,
+		statefulSets: statefulSets,
+		recorder:     recorder,
+	}
+
+}
+
+func (c *pilotControl) clusterPods(cluster *v1alpha1.CassandraCluster) ([]*v1.Pod, error) {
+	var clusterPods []*v1.Pod
+	allPods, err := c.pods.Pods(cluster.Namespace).List(labels.Everything())
+	if err != nil {
+		return clusterPods, err
+	}
+	for _, pod := range allPods {
+		podControlledByCluster, err := controllers.PodControlledByCluster(
+			cluster,
+			pod,
+			c.statefulSets,
+		)
+		if err != nil {
+			return clusterPods, err
+		}
+		if !podControlledByCluster {
+			continue
+		}
+		clusterPods = append(clusterPods, pod)
+	}
+	return clusterPods, nil
+}
+
+func (c *pilotControl) createOrUpdatePilot(cluster *v1alpha1.CassandraCluster, pod *v1.Pod) error {
+	desiredPilot := PilotForCluster(cluster, pod)
+	client := c.naviClient.NavigatorV1alpha1().Pilots(desiredPilot.GetNamespace())
+	lister := c.pilots.Pilots(desiredPilot.GetNamespace())
+	existingPilot, err := lister.Get(desiredPilot.GetName())
+	if k8sErrors.IsNotFound(err) {
+		_, err = client.Create(desiredPilot)
+		return err
+	}
+	if err != nil {
+		return err
+	}
+	err = util.OwnerCheck(existingPilot, cluster)
+	if err != nil {
+		return err
+	}
+	existingPilot = existingPilot.DeepCopy()
+	existingPilot.Status = v1alpha1.PilotStatus{}
+	desiredPilot = existingPilot.DeepCopy()
+	desiredPilot = updatePilotForCluster(cluster, pod, desiredPilot)
+	if !reflect.DeepEqual(desiredPilot, existingPilot) {
+		_, err = client.Update(desiredPilot)
+	}
+	return err
+}
+
+func (c *pilotControl) syncPilots(cluster *v1alpha1.CassandraCluster) error {
+	pods, err := c.clusterPods(cluster)
+	if err != nil {
+		return err
+	}
+	for _, pod := range pods {
+		err = c.createOrUpdatePilot(cluster, pod)
+		if err != nil {
+			return err
+		}
+	}
+	return err
+}
+
+func (c *pilotControl) Sync(cluster *v1alpha1.CassandraCluster) error {
+	err := c.syncPilots(cluster)
+	if err != nil {
+		return err
+	}
+	// TODO: Housekeeping. Remove pilots that don't have a corresponding pod.
+	return nil
+}
+
+func PilotForCluster(cluster *v1alpha1.CassandraCluster, pod *v1.Pod) *v1alpha1.Pilot {
+	pilot := &v1alpha1.Pilot{}
+	pilot.SetOwnerReferences(
+		[]metav1.OwnerReference{
+			util.NewControllerRef(cluster),
+		},
+	)
+	return updatePilotForCluster(cluster, pod, pilot)
+}
+
+func updatePilotForCluster(
+	cluster *v1alpha1.CassandraCluster,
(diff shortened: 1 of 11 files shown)


wallrj:
This isn't working yet. 
Forgot that the ownerreferences are to the statefulset not to the cassandracluster.
Looking at whether it's possible to set additional ownerreferences  from the statefulset template.

wallrj:
I've updated this PR against master. 
There are two other PRs stacked on top of this (#142, #162)
But I've marked those as WIP to avoid confusion.
Ready for review.
/cc @munnerz

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		_, found := expectedPilotNames[pilot.Name]
+		if !found {
+			err := client.Delete(pilot.Name, nil)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
```
munnerz: So I'm not too sure if we should be doing this right now (deleting 'unused' pilot resources). It's difficult to define unused, and in fact, the lack of a pod does not immediately imply that the Pilot resource is no longer in use. I'd rather do something along the lines of waiting N seconds since the last heartbeat (provided no corresponding pod exists) before deleting.

Right now the Elasticsearch controller doesn't actually clean up old Pilots, it just cleans up old StatefulSets (i.e. ones no longer specified as a nodepool on the ElasticsearchCluster resource). Should we make these the same? Consistency in this sort of thing probably makes it easier to reason about. At some point I think we can refactor the Pilot control loop to make it mostly generic between DB types.

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return err
+	}
+	err = util.OwnerCheck(existingPilot, cluster)
+	if err != nil {
+		return err
+	}
+	desiredPilot = existingPilot.DeepCopy()
+	updatePilotForCluster(cluster, pod, desiredPilot)
+	_, err = client.Update(desiredPilot)
```
munnerz: So from what I can see here, this will cause Navigator to go into a sync loop, continuously updating the Pilot resource. We don't ever check if the Pilot is already reconciled (e.g. it's spec is up to date), so we end out updating to contain the same spec as is already there. This will still however increment the ResourceVersion, thus causing this loop to be triggered once again (and so on).

There are a couple of ways (that I can think of) to deal with this:

* Perform a `reflect.DeepEqual` on the `pilot.Spec` before calling Update. This has the downside of being inefficient (DeepEqual is an expensive call), however it is accurate and easy to do.

* Write a hash of the spec into the annotations of the Pilot, so that we can quickly compare the old and new hashes.

* Manually compare each field, but this is tedious and error prone.

Can you think of anything else here?

munnerz: Additionally, we need to update the strategy.go file for Cassandra resources to not perform updates to the `cassandracluster.status` block when plain Update is called (and instead, `UpdateStatus` should be used if the Status block needs updating).

Right now, each time Update is called we will wipe out all fields in Status, causing Pilots and Navigator to fight with each other indefinitely (causing more loops)

pkg/controllers/util_api.go
```diff
@@ -0,0 +1,33 @@
+package controllers
+
+import (
+	apiv1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	appslisters "k8s.io/client-go/listers/apps/v1beta1"
+)
+
+func PodControlledByCluster(
```
munnerz: Can we also update the Elasticsearch controller to use this function too? This version doesn't depend on any types like *ElasticsearchCluster like the one the the controller currently uses

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	err = c.removeUnusedPilots(cluster)
+	return err
+}
+
+func PilotForCluster(cluster *v1alpha1.CassandraCluster, pod *v1.Pod) *v1alpha1.Pilot {
+	pilot := &v1alpha1.Pilot{}
+	ownerRefs := pilot.GetOwnerReferences()
+	ownerRefs = append(ownerRefs, util.NewControllerRef(cluster))
+	pilot.SetOwnerReferences(ownerRefs)
```
munnerz: Don't need so many steps to do this bit - we aren't mutating an existing Pilot, but creating a new one instead (so don't need to use function accessors)

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func updatePilotForCluster(
+	cluster *v1alpha1.CassandraCluster,
+	pod *v1.Pod,
+	pilot *v1alpha1.Pilot,
+) *v1alpha1.Pilot {
+	pilot.SetName(pod.GetName())
+	pilot.SetNamespace(cluster.GetNamespace())
+	pilot.SetLabels(util.ClusterLabels(cluster))
```
munnerz: We should be careful not to override any user provided labels, or labels that may have been added by the pilot (we don't do this right now, but we might do at some point)

pkg/controllers/cassandra/pilot/pilot.go
wallrj: > Write a hash of the spec into the annotations of the Pilot, so that we can quickly compare the old and new hashes.

I've gone with that option. I found the ES hashing code and adapted that. Also added labels to the hash since the controller updates those too.

wallrj: > Right now, each time Update is called we will wipe out all fields in Status, causing Pilots and Navigator to fight with each other indefinitely (causing more loops)

I don't think that's the case. I'm taking the latest Pilot from the lister and then updating labels and in future, spec Not replacing Pilot.Status.

pkg/controllers/cassandra/pilot/pilot.go
wallrj: > So I'm not too sure if we should be doing this right now (deleting 'unused' pilot resources). It's difficult to define unused, and in fact, the lack of a pod does not immediately imply that the Pilot resource is no longer in use. I'd rather do something along the lines of waiting N seconds since the last heartbeat (provided no corresponding pod exists) before deleting.

But what's the bug that the heartbeat / delay would fix?
If there is a bug, then I can add the heartbeat / delay in a followup branch.
But for now, removing unused Pilot resources seems like good housekeeping.

pkg/controllers/cassandra/pilot/pilot.go
wallrj: Done.

pkg/controllers/cassandra/pilot/pilot.go
wallrj: Done.

pkg/controllers/util_api.go
wallrj: Done.

pkg/controllers/cassandra/pilot/pilot.go
munnerz: The heartbeat is a suggestion for how it can be handled, but as you say, should be in a follow up.

I think that we can't say the lack of pod = the pilot doesn't exist. In the ES controller for example, the 'master' Pilot uses Pilot resources to determine whether a cluster is in a state ready to scale down for example. This scenario, as far as I can tell, will cause failures if we auto-delete Pilots if no Pod exists:

* User requests scale down of cluster
* Navigator updates a Pilot resource to 'decommission' it
* Because the user is a sadist, or because *something* happened, the pod that is being decommissioned gets deleted.
* This *should* cause the StatefulSet controller to re-create the pod, really, as the Pilot is *not* finished decommissioning so it needs to come back in order to finish
* However, because we immediately deleted the Pilot resource upon the Pod disappearing, the 'knowledge' of those remaining documents/indices left on that node (i.e. the `pilot.status.elasticsearch.documentCount` field) has been lost.
* So now our master Pilot is not aware of those extra documents, nor the fact that a Pilot is being decommissioned either (as that knowledge has also been lost).
* At this point we've reached a bit of a weird state - especially if we consider that the *only* state about the state of the world is stored in the k8s api and anything in-memory should not be relied upon

I do agree we need to do house keeping, but I'm thinking we should consider *how* this is done carefully as 'unused' is a tricky term. Additionally, given the Pilot resource in the API is one of the users primary points for debugging issues with their clusters, by deleting the Pilots so frequently we are also destroying a lot of useful debugging information for users. I'd much prefer the behaviour between all DB types to be consistent - the first steps in debugging the ES cluster should be the same as Cassandra, and vice versa.

munnerz: ^ my only outstanding/current issue with this PR before I'm happy to merge 😄 .

Let me know if you disagree, but any change we make here needs to be consistent with ES too given this touches a user-facing resource (the Pilot resource)

wallrj: It seems to me that in the scenario above,

* Accidentally deleted pod gets restarted by StatefulSet Controller.
* Pilot process waits for appearance of Pilot
* Pilot starts ES sub-process
* Pilot re-updates the Pilot status with document count 0
* navigator controller waits for document count 0 and then decrements the Statefulset replicas count 
* and removes the drained pod and corresponding pilot.

But I agree that we should keep ES and Cass in sync, so I'll remove the Pilot cleanup code for now.

munnerz:
I restarted the build infra workers

/retest

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -0,0 +1,182 @@
...
+	return pilot
+}
+
+func ComputeHash(p *v1alpha1.Pilot) uint32 {
+	hashVar := []interface{}{
+		p.Spec,
+		p.ObjectMeta,
+		p.Labels,
+	}
+	hasher := fnv.New32()
```
munnerz: Can this be pulled out to be a global var (to save calling New32 each time)?

Happy to be a follow up as I'll be making similar changes in ES

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -0,0 +1,182 @@
...
+		return clusterPods, err
+	}
+	for _, pod := range allPods {
+		podControlledByCluster, err := controllers.PodControlledByCluster(
+			cluster,
+			pod,
+			c.statefulSets,
+		)
+		if err != nil {
+			return clusterPods, err
```
munnerz: Should probably `return nil, err` here instead of a partial list of pods.

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -0,0 +1,182 @@
...
+	if k8sErrors.IsNotFound(err) {
+		_, err = client.Create(desiredPilot)
+		return err
+	}
+	if err != nil {
+		return err
+	}
+	err = util.OwnerCheck(existingPilot, cluster)
+	if err != nil {
+		return err
```
munnerz: TODO: should there be some way to detect that this function has failed because the existing pilot is owned by another cluster? (not important for merge)

Issue #154, opened 2017-11-28
Ready and Liveness probes fail because timeout is too short


wallrj:
/kind bug

The 5s timeout is too short.
In my test cluster, the `/ready-probe.sh` script takes up to 10s.

PR #155, opened 2017-11-28
Increase the timeout for Ready and Liveness probes

wallrj:
* Increase the timeout for Ready and Liveness probes.
* Use a lower failure threshold for readiness probe, so that the CQL service bypasses a dead node faster.
* Use a larger success threshold for readiness probe, so that traffic isn't redirected to a node until it has been consistently stable.
* Fix a bug in the simulation of a dead node. I was stopping all processes in the container, including the readiness and liveness probes themselves, if that happened to be running.
* Wait until both nodes in the 2-node cluster are *ready*, before commencing the dead node bypass test.

Fixes: #154
Fixes: #160

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cassandra/nodepool/resource.go
+++ b/pkg/controllers/cassandra/nodepool/resource.go
@@ -62,15 +62,20 @@ func StatefulSetForCluster(
 										// It can perform similar ready probe.
 										Command: []string{
 											"/usr/bin/timeout",
-											"5",
+											"10",
 											"/ready-probe.sh",
 										},
 									},
 								},
-								InitialDelaySeconds: 15,
+								// Test logs show that Cassandra begins
+								// listening for CQL connections ~45s after startup.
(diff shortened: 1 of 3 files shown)


pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -84,15 +89,21 @@ func StatefulSetForCluster(
 									Exec: &apiv1.ExecAction{
 										Command: []string{
 											"/usr/bin/timeout",
-											"5",
+											"10",
```
munnerz: Why do we use timeout here as well as setting a timeout on the probe itself?

munnerz:
Can we add a TODO to this part of code re: timeouts to remind us why we've done this and to change it once fixed or something for future reference? Can be a follow up PR though :)

/lgtm
/approve

PR #159, opened 2017-11-30
WIP: Retry CQL connection from inside a single pod

wallrj:
To avoid having so many uninteresting pod events in the cluster-info after the tests finish.

**Release note**:
```release-note
NONE
```


munnerz:
Looks like this is passing - is there any reason why this is still WIP @wallrj?

Issue #161, opened 2017-11-30
RBAC policy prevents Pilot from listing Pilot resources


wallrj:
The cassandra pilot currently runs as default service account in its namespace and this does not have permission to list or update status of pilot resources.

Part of: #140 

/kind bug

PR #162, opened 2017-11-30
Pilot service account and RBAC policy

wallrj:
* Add a service account for Pilots
* and a corresponding RBAC policy which allows pilots to update their pilot resources.

Fixes: #161
**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cassandra/testing/testing.go
+++ b/pkg/controllers/cassandra/testing/testing.go
@@ -4,13 +4,17 @@ import (
 	"testing"
 
 	navinformers "github.com/jetstack/navigator/pkg/client/informers/externalversions"
+	rbacv1 "k8s.io/api/rbac/v1beta1"
 
 	"github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra/nodepool"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra/pilot"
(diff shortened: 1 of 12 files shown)


munnerz:
Is this meant to be merged? It seems to do a lot more than just add a service account and RBAC policy 🤔

pkg/controllers/cassandra/testing/testing.go
```diff
@@ -174,6 +213,78 @@ func (f *Fixture) AssertServicesLength(l int) {
...
+func (f *Fixture) RoleBindings() *rbacv1.RoleBindingList {
+	roleBindings, err := f.k8sClient.
+		RbacV1beta1().
+		RoleBindings(f.Cluster.Namespace).
+		List(metav1.ListOptions{})
+	if err != nil {
+		f.t.Fatal(err)
+	}
+	return roleBindings
+}
```
munnerz: For a future PR - but it'd be great if we can make these sorts of things generic between controllers.

If we define a `RoleBindingsFixture` struct, and embed it within the `Fixture` struct for each controller we can plug and play different test fixture pieces potentially.

Something to think about 😄

munnerz:
From looking at the logs from the 1.9 failure, it seems there may be a bug in either genericpilot or Elasticsearch pilot that is causing it not to sync as often as it should (logs: https://storage.googleapis.com/jetstack-logs/pr-logs/pull/jetstack_navigator/162/navigator-e2e-v1-9/90/artifacts/test-elasticsearchcluster-1516211071-11552/dump.

The pilot wasn't being synced for >1m, causing it to not update the document count on time.

Issue #169, opened 2017-12-05
Have the pilot update the status of its PilotResource


wallrj:
Part of #140 

/kind bug

wallrj:
Trying the Jolokia javaagent in order to access database status information via a REST API rather than having to use nodetool.

```
wget --quiet -O - http://localhost:8778/jolokia/list?maxDepth=1 | jq
wget https://github.com/rhuss/jolokia/releases/download/v1.3.7/jolokia-1.3.7-bin.tar.gz
tar xf jolokia-1.3.7-bin.tar.gz
-javaagent:/jolokia-1.3.7/agents/jolokia-jvm.jar
wget --quiet -O - http://localhost:8778/jolokia/read/org.apache.cassandra.db:type=StorageService/LiveNodes | jq .
```

See:
 * https://engineeringblog.yelp.com/2016/06/monitoring-cassandra-at-scale.html
 * https://github.com/apache/cassandra/blob/cassandra-3.11/src/java/org/apache/cassandra/service/StorageServiceMBean.java
 * https://jolokia.org/reference/html/protocol.html
 * http://blog.tomas.cat/en/2013/04/12/monitoring-cassandra-relevant-data-should-be-watched-and-how-send-it-graphite/
 * https://github.com/rhuss/jolokia/issues/256
 * https://medium.com/@mlowicki/cassandra-metrics-and-their-use-in-grafana-1f0dc33f9cca
 * https://medium.com/@mlowicki/monitoring-cassandra-a7fde5b2de9c

munnerz:
I've taken a quick look at Jolokia & nodetool, although as you can imagine it's not something I've read much about before.

From what I can see, nodetool doesn't support output formats when you run a CLI command, so it'll take some nasty parsing to use it instead of some programmatic API that Jolokia provides.

I'm kind of hesitant however to set up/run Jolokia within each pod, simply because it's not something that we're familiar with yet, and given it's quite specific to Cassandra I think we could do with chatting to some experts to make sure that this *is* in fact the defacto for it. Is there any way we can introduce the Pilot resource type without this for now?

If this is something we can't do without, I think we should sync up next week and get on a call with someone that knows Cassandra well to verify our decisions 😄 wdyt?

wallrj:
Fixed in #263

PR #170, opened 2017-12-06
Jolokia based readiness and liveness probes

wallrj:
* Load Jolokia agent when Cassandra node starts.
* A golang client for getting local Cassandra node status
* Linked to the Pilot readiness and liveness health functions.

You can see the source of `nodetool status`, which this replaces, here:
 * https://git1-us-west.apache.org/repos/asf?p=cassandra.git;a=blob;f=src/java/org/apache/cassandra/tools/nodetool/Status.java;h=f7bb1e52a786fe40ec9a2d6d866618c814562b3f;hb=refs/heads/cassandra-3.11

Part of #169 

**Release note**:
```release-note
NONE
```


wallrj:
Tests failed due to Helm upgrade on our build infrastructure.
 * https://github.com/kubernetes/helm/issues/3382

We've downgraded helm, so let's try again.

/retest

munnerz:
/retest

(1.8 was already running, so wasn't restarted upon your first retest)

pkg/cassandra/nodetool/client/client.go
```diff
@@ -0,0 +1,89 @@
...
+	MovingNodes      []string
+	LocalHostId      uuid.UUID
+}
+
+type Interface interface {
+	StorageService() (*StorageService, error)
+}
+
+type client struct {
+	StorageServiceURL *url.URL
```
munnerz: Does this field need to be exported, given it's a non-exported struct?

pkg/cassandra/nodetool/client/client.go
```diff
@@ -0,0 +1,89 @@
...
+	client            *http.Client
+}
+
+var _ Interface = &client{}
+
+func New(baseURL *url.URL, c *http.Client) *client {
+	storageServiceURL, err := url.Parse(StorageServicePath)
+	if err != nil {
+		panic(err)
```
munnerz: Better to return an error rather then panic - we can then deal with that error accordingly 'upstream'

pkg/cassandra/nodetool/client/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+type StorageService struct {
+	HostIdMap        map[string]uuid.UUID
+	LiveNodes        []string
+	UnreachableNodes []string
+	LeavingNodes     []string
+	JoiningNodes     []string
+	MovingNodes      []string
+	LocalHostId      uuid.UUID
+}
```
munnerz: From what I can tell, this struct is deseralised into from JSON? If so, please add `json:""` struct tags so it's explicit. I guess `uuid.UUID` defines its own Marshal/Unmarshal methods?

pkg/cassandra/nodetool/client/client.go
```diff
@@ -0,0 +1,89 @@
...
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to read response body. '%s'", err)
+	}
+
+	out := &JolokiaResponse{}
+
+	err = json.Unmarshal(body, out)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to unmarshal JSON response. '%s'", err)
```
munnerz: returned errors (as opposed to printed errors) should not start with a caps according to go style guides/conventions.

Generally it'd be on that call to `glog.Infof` where you would prepend whatever error with a nicely formatted sentence, e.g.:

```go
glog.Infof("Error obtaining storage service: %v", err)
```

Please also use a colon at the end of your extra informational message to make it clear another error follows, e.g.:

```
return nil, fmt.Errorf("failed to unmarshal JSON response: %v", err)
```

pkg/cassandra/nodetool/client/client.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if out.Value == nil {
+		return nil, fmt.Errorf("The response had an empty Jolokia value. '%#v'", out)
```
munnerz: As out.Value == nil, and out only contains a field named Value, surely printing this here will print nothing?

munnerz: (also: 'The' -> 'the')

pkg/cassandra/nodetool/client/client.go
```diff
@@ -0,0 +1,89 @@
+package client
```
munnerz: Can we get rid of this sub-package and just absorb its functionality into `pkg/cassandra/nodetool`? I'm not too sure what the difference between these packages is, it seems they are both clients? Can this be folded into a single client? 😄

pkg/cassandra/nodetool/nodetool.go
```diff
@@ -0,0 +1,170 @@
...
+type NodeMap map[string]*Node
+
+func (nm NodeMap) LocalNode() *Node {
+	for _, node := range nm {
+		if node.Local {
+			return node
+		}
+	}
+	return nil
+}
```
munnerz: Might be worth adjusting this method to return an error if the LocalNode isn't found? Else we might accidentally get some nil pointer exceptions further down the line.

pkg/cassandra/nodetool/nodetool.go
```diff
@@ -0,0 +1,170 @@
...
+
+type Interface interface {
+	Status() (NodeMap, error)
+}
+
+type tool struct {
+	client client.Interface
```
munnerz: Ah, so the `nodetool` package depends upon the `client` package. It'd be great if we can get rid of this `tool` struct, and instead just have a `client` and then update the above `Interface` to include whatever is defined in `client`. Perhaps I'm missing something/perhaps I don't understand why you've done this in the first place.

pkg/cassandra/nodetool/nodetool.go
```diff
@@ -0,0 +1,170 @@
...
+			Status: NodeStatusUnknown,
+			State:  NodeStateNormal,
+		}
+	}
+
+	liveNodes := sets.NewString(ssInfo.LiveNodes...)
+	unreachableNodes := sets.NewString(ssInfo.UnreachableNodes...)
+	if setsIntersect(liveNodes, unreachableNodes) {
+		return nil, fmt.Errorf(
+			"The sets of live and unreachable nodes should not intersect. "+
```
munnerz: drop caps from the start of this error.
you might then want to do the same for Live/Unreachable but idm. convention does not dictate anything beyond the first char

pkg/cassandra/nodetool/nodetool.go
```diff
@@ -0,0 +1,170 @@
...
+
+		switch {
+		case leavingNodes.Has(host):
+			node.State = NodeStateLeaving
+		case joiningNodes.Has(host):
+			node.State = NodeStateJoining
+		case movingNodes.Has(host):
+			node.State = NodeStateMoving
+		default:
+			node.State = NodeStateNormal
```
munnerz: Is there ever any case where NodeStateUnknown would be returned? Can't see it here.

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -62,6 +63,29 @@ func StatefulSetForCluster(
 					},
 					InitContainers: []apiv1.Container{
 						pilotInstallationContainer(&cluster.Spec.NavigatorClusterConfig.PilotImage),
+						apiv1.Container{
+							Name:            "install-agent-bond",
+							Image:           "fabric8/java-alpine-openjdk8-jre:latest",
```
munnerz: Instead of using an image that contains a full jre just to grab a jar file out of it, can we include this one jar in the pilot-cassandra image and adjust this initContainer to use this? By both initContainers referencing the same image, it saves it being pulled multiple times 😄

pkg/pilot/cassandra/v3/options.go
```diff
@@ -66,6 +72,12 @@ func (o *PilotOptions) AddFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Master, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	flags.DurationVar(&o.ResyncPeriod, "resync-period", defaultResyncPeriod, "Re-sync period for control loops operated by the pilot")
 	flags.StringVar(&o.ConfigDir, "config-dir", defaultConfigDir, "Base directory for additional Pilot configuration")
+	flags.StringVar(
+		&o.JolokiaURL,
+		"jolokia-url",
+		"",
+		"The base URL of the Jolokia REST API server",
+	)
```
munnerz: Can you collapse this into a single line for consistency with other lines in this fn?

pkg/pilot/cassandra/v3/options.go
```diff
@@ -85,6 +97,11 @@ func (o *PilotOptions) Complete() error {
 	}
 	o.sharedInformerFactory = informers.NewFilteredSharedInformerFactory(o.navigatorClientset, o.ResyncPeriod, o.GenericPilotOptions.PilotNamespace, nil)
 
+	u, err := url.Parse(o.JolokiaURL)
+	if err != nil {
+		return err
+	}
```
munnerz: What happens if JolokiaURL is not set?

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -67,12 +68,28 @@ func (p *Pilot) syncFunc(pilot *v1alpha1.Pilot) error {
...
 }
 
+func (p *Pilot) ReadinessCheck() error {
+	return localNodeUpAndNormal(p.nodeTool)
+}
+
 func (p *Pilot) LivenessCheck() error {
-	glog.V(2).Infof("liveness status: %q", "ok")
-	return nil
+	return localNodeUpAndNormal(p.nodeTool)
```
munnerz: The liveness check here probably shouldn't rely on the cassandra node being up and normal. If a node is recovering, or still starting up, the *readiness* probe should fail but liveness should pass. If a liveness probe fails, Kubernetes will begin restarting that pod, which can quite easily cause the pod to never start as a result of constant CrashLoopBackOff

pkg/pilot/genericpilot/probe/listen.go
```diff
@@ -16,8 +18,12 @@ func (l *Listener) Listen() error {
 		fmt.Sprintf(":%d", l.Port),
 		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			if err := l.Check(); err != nil {
+				glog.Errorf("Error '%s', returned by Check function for probe on port %d", err, l.Port)
```
munnerz: Can you re-arrange this so `err` is at the end? `failed running Check function for probe on port %d: %v` (also no quotation marks around the error message)

pkg/pilot/genericpilot/probe/listen.go
```diff
@@ -16,8 +18,12 @@ func (l *Listener) Listen() error {
...
 				w.WriteHeader(http.StatusInternalServerError)
-				w.Write([]byte(err.Error()))
+				_, err := w.Write([]byte(err.Error()))
+				if err != nil {
+					glog.Errorf("Error '%s' while writing error message for probe on port %d", err, l.Port)
```
munnerz: Again re-arrange to have err at the end and no quotes

pkg/cassandra/nodetool/client/client.go
wallrj: I'll leave this. It's parsing a string constant defined above so the error will only ever be seen by us. Similar to `resource.MustParse`: https://godoc.org/k8s.io/apimachinery/pkg/api/resource#MustParse

pkg/cassandra/nodetool/nodetool.go
wallrj: I'll leave it for now.

pkg/cassandra/nodetool/client/client.go
munnerz: Ah - did not notice that. Can we move this into an `init` function perhaps, and make `storageServiceURL` a package scoped var? These execute at the time the package is initially loaded, and there can be as many of them as you want. All you should need to do is add:

```
var storageServiceURL *url.URL
func init() {
    var err error
    storageServiceURL, err = url.Parse(StorageServicePath)
    if err != nil {
        panic(err)
    }
}
```

It'll mean the error occurs at the time the program starts, instead of further along in execution (making it a lot more obvious there's a problem with the string, as the binary itself won't start at all with it). Also saves parsing the URL multiple times 😄

pkg/cassandra/nodetool/nodetool.go
wallrj: My idea was to separate out the Jolokia specific stuff into a separate package....it'd be better named `github.com/jetstack/navigator/pkg/cassandra/jolokia`.
* This would allow the interactions with Jolokia to be tested separately.
* And the `nodetool.Interface` could be tested using a fake implementation of `client.Interface` which returns canned responses.
* And if Jolokia doesn't work out, we could add a client that shells out to `nodetool` and parses its responses.

But I haven't got that far yet. I suggest I refactor in followup branches as I discover more about the requirements.

pkg/cassandra/nodetool/nodetool.go
wallrj: Yeah, if a node isn't in either of LiveNodes or Unreachable nodes.
The `?` case here: https://git1-us-west.apache.org/repos/asf?p=cassandra.git;a=blob;f=src/java/org/apache/cassandra/tools/nodetool/Status.java;h=f7bb1e52a786fe40ec9a2d6d866618c814562b3f;hb=refs/heads/cassandra-3.11#l155

pkg/pilot/cassandra/v3/options.go
wallrj: I'd like to be able to make it a required flag, but the flags package doesn't support that. 
I added a default.

pkg/cassandra/nodetool/nodetool.go
munnerz: 👍 makes sense. I much prefer `github.com/jetstack/navigator/pkg/cassandra/jolokia` too. It's nice and obvious what a `github.com/jetstack/navigator/pkg/cassandra/jolokia.Client` does too (`jolokia.Client` is clearly a jolokia client 😄)

pkg/cassandra/nodetool/nodetool.go
munnerz: So I see we have `NodeStatusUnknown` as well as `NodeStateUnknown`, although `NodeStateUnknown` is not used in this `for` loop, hence my question (as it seems that there are no cases where `node.State` is set to `NodeStateUnknown`).

pkg/cassandra/nodetool/nodetool.go
```diff
@@ -0,0 +1,170 @@
...
+	NodeStatusUnknown NodeStatus = "Unknown"
+	NodeStatusUp      NodeStatus = "Up"
+	NodeStatusDown    NodeStatus = "Down"
+)
+
+type Node struct {
+	Host   string
+	ID     uuid.UUID
+	State  NodeState
+	Status NodeStatus
```
munnerz: What's the difference between State and Status? I see they have different values (above), but not sure what each of them mean. A brief comment here would help.

pkg/cassandra/nodetool/nodetool.go
```diff
@@ -0,0 +1,170 @@
...
+	for _, node := range nm {
+		if node.Local {
+			return node
+		}
+	}
+	return nil
+}
+
+type Interface interface {
+	Status() (NodeMap, error)
```
munnerz: Could we rename `Status` to `Nodes`? Status is becoming quite an overloaded term in this package.

pkg/cassandra/nodetool/nodetool.go
munnerz: Yeah from re-reading this whole function it appears we should set `node.State = NodeStateNormal` *iff* the node is a member of `liveNodes`? Otherwise we should set this to unknown (which seems equiv. to 'unreachable'? Unless we need that as an *extra* state).

pkg/cassandra/nodetool/nodetool.go
```diff
@@ -0,0 +1,170 @@
...
+	NodeStateMoving  NodeState = "Moving"
+)
+
+type NodeStatus string
+
+const (
+	NodeStatusUnknown NodeStatus = "Unknown"
+	NodeStatusUp      NodeStatus = "Up"
+	NodeStatusDown    NodeStatus = "Down"
+)
```
munnerz: Will these strings be surfaced via our API (i.e. on a pilot.status field?).

If so, we should earmark them to move into `pkg/apis/navigator/v1alpha1`. Not important for merge though as afaict, they are not being used on API types currently.

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -179,6 +191,16 @@ func StatefulSetForCluster(
...
+								{
+									Name: "JVM_OPTS",
+									Value: fmt.Sprintf(
+										"-javaagent:%s/jolokia.jar=host=%s,port=%d,agentContext=%s",
+										sharedVolumeMountPath,
+										jolokiaHost,
+										jolokiaPort,
+										jolokiaContext,
+									),
+								},
```
munnerz: This is okay for now, but I think this environment variable should be set/managed by Pilot in future. We've got a similar question/issue coming up in the ES controller, as we need to see the heap size (via env var) too.

pkg/pilot/cassandra/v3/options.go
munnerz: 👍 yep, flags package in golang doesn't support required as there is a mantra of "no flag is required", and that defaults should be used instead.

I'm not sure I am convinced by the mantra, but in this case, the use of a default works well. I think it's rare that a user would ever want to override this, given the Pilot itself is what configures Jolokia anyway so it is better informed on the URL to use.

pkg/controllers/cassandra/nodepool/resource.go
munnerz: (line comment went weird - this is specifically around setting the JAVA_OPTS env var)

pkg/cassandra/nodetool/nodetool.go
wallrj: I've added comments in the source code.

pkg/cassandra/nodetool/nodetool.go
wallrj: I'm pretty sure I'm exactly replicating the logic from nodetool.

```
 153         if (liveNodes.contains(endpoint)) status = "U";
 154         else if (unreachableNodes.contains(endpoint)) status = "D";
 155         else status = "?";
 156         if (joiningNodes.contains(endpoint)) state = "J";
 157         else if (leavingNodes.contains(endpoint)) state = "L";
 158         else if (movingNodes.contains(endpoint)) state = "M";
 159         else state = "N";
```

The reason for `NodeStateUnknown` is that that I wanted a sensible zero-value.
You're right that the `nodetool.Status` function will never return it, but none of the other states make sense as zero value.

pkg/cassandra/nodetool/nodetool.go
wallrj: Good idea, but again, I wanted to try and match nodetool as closely as possible.
This is intended to give a structured version of the `nodetool status` output.
Let's leave it as is for now and I can refactor later when I've got a better feeling for what operations we need to perform.

munnerz: 👍

PR #175, opened 2017-12-13
Dump namespace info/logs on test failure to artifacts directory

munnerz:
/cc @wallrj 

```release-note
NONE
```

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -93,53 +93,49 @@ function fail_test() {
 }
 
 function test_elasticsearchcluster() {
+    local namespace="${1}"
     echo "Testing ElasticsearchCluster"
-    local USER_NAMESPACE="test-elasticsearchcluster-${TEST_ID}"
-    kubectl create namespace "${USER_NAMESPACE}"
+    kubectl create namespace "${namespace}"
     if ! kubectl get esc; then
         fail_test "Failed to use shortname to get ElasticsearchClusters"
     fi
     # Create and delete an ElasticSearchCluster
     if ! kubectl create \
-            --namespace "${USER_NAMESPACE}" \
+            --namespace "${namespace}" \
             --filename "${SCRIPT_DIR}/testdata/es-cluster-test.yaml"; then
         fail_test "Failed to create elasticsearchcluster"
     fi
     if ! kubectl get \
-            --namespace "${USER_NAMESPACE}" \
+            --namespace "${namespace}" \
             ElasticSearchClusters; then
         fail_test "Failed to get elasticsearchclusters"
     fi
     if ! retry kubectl get \
-         --namespace "${USER_NAMESPACE}" \
+         --namespace "${namespace}" \
          service es-test; then
         fail_test "Navigator controller failed to create elasticsearchcluster service"
     fi
-    if ! retry kube_event_exists "${USER_NAMESPACE}" \
+    if ! retry kube_event_exists "${namespace}" \
          "navigator-controller:ElasticsearchCluster:Normal:SuccessSync"
     then
         fail_test "Navigator controller failed to create SuccessSync event"
     fi
     if ! retry TIMEOUT=300 stdout_gt 0 kubectl \
-         --namespace "${USER_NAMESPACE}" \
+         --namespace "${namespace}" \
          get pilot \
          "es-test-mixed-0" \
          "-o=go-template={{.status.elasticsearch.documents}}"
     then
         fail_test "Elasticsearch pilot did not update the document count"
     fi
-    if ! kubectl delete \
-            --namespace "${USER_NAMESPACE}" \
-            ElasticSearchClusters \
-            --all; then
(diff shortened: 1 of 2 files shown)


hack/libe2e.sh
```diff
@@ -101,3 +99,49 @@ function stdout_gt() {
...
     return 1
 }
+
+function dump_debug_logs() {
+    local namespace="${1}"
+    if [ "${namespace}" != "" ]; then
+        local output_dir="$(pwd)/_artifacts/${namespace}/dump.txt"
+        echo "Dumping cluster state to ${output_dir}"
+        mkdir -p "${output_dir}"
+        kubectl cluster-info dump --namespaces "${namespace}" > "${output_dir}/dump.txt" || true
```
wallrj: You can do `kubectl cluster-info dump --namespaces default,kube-system --output-directory=/path/to/cluster-state` which will output separate yaml and log files.

hack/libe2e.sh
```diff
@@ -101,3 +99,49 @@ function stdout_gt() {
     fi
     return 1
 }
+
+function dump_debug_logs() {
+    local namespace="${1}"
+    if [ "${namespace}" != "" ]; then
```
wallrj: `namespace` should never be empty, right? Why check for it?

hack/libe2e.sh
```diff
@@ -101,3 +99,49 @@ function stdout_gt() {
...
+            | ignore_expected_controller_errors
+    then
+        fail_test "Unexpected errors in controller logs" "${namespace}"
+    fi
+    if kubectl logs -c apiserver -l app=navigator,component=apiserver \
+            | egrep '^E[0-9]{4} '
+    then
+        fail_test "Unexpected errors in apiserver logs" "${namespace}"
+    fi
+}
```
wallrj: Ok, you win :-) Let's just delete this test now that we have full logs for each test run.

hack/e2e.sh
```diff
@@ -95,79 +95,53 @@ function fail_test() {
...
+    fail_and_exit "${USER_NAMESPACE}"
+fi
 
-exit $FAILURE_COUNT
+if ! kubectl delete \
+        --namespace "${USER_NAMESPACE}" \
+        ElasticSearchClusters \
+        --all; then
+    fail_test "Failed to delete elasticsearchcluster" "${USER_NAMESPACE}"
+fi
```
wallrj: This was originally part of the E2E tests....to ensure that the API server allowed our resources to be deleted.
But that's now covered by the deletion of the namespace.
I'd be inclined to add `ElasticSearchClusters` to the list of resources that we explicitly delete in `kube_delete_namespace_and_wait`.

hack/e2e.sh
```diff
@@ -95,79 +95,53 @@ function fail_test() {
...
+if ! kube_delete_namespace_and_wait "${USER_NAMESPACE}"; then
+    fail_test "Failed to delete test namespace" "${USER_NAMESPACE}"
+fi
```
wallrj: I think this should be simply `kube_delete_namespace_and_wait "${USER_NAMESPACE}"` which will cause the e2e.sh script to exit if it exits with > 0.

wallrj:
Ah, sorry, my fault for suggesting adding the ElasticSearchCluster to the list of resources to be deleted.
That won't work with the initial namespace deletion because navigator api server isn't running yet.
Maybe it isn't necessary.

munnerz:
Trying to pick up flakes

/retest

hack/libe2e.sh
```diff
@@ -122,3 +120,23 @@ function stdout_gt() {
     fi
     return 1
 }
+
+function dump_debug_logs() {
+    local namespace="${1}"
+    local output_dir="$(pwd)/_artifacts/${namespace}"
+    echo "Dumping cluster state to ${output_dir}"
+    mkdir -p "${output_dir}"
+    kubectl cluster-info dump --namespaces "${namespace}" > "${output_dir}/dump.txt" || true
```
wallrj: ❓ I mentioned it in previous review, but it might be good to use the `--output-directory` option here, which will result in a directory with separate yaml files for each of the resource types and separate log files for each of the pods.

wallrj:
Ah. Sorry @munnerz 

I didn't mean for it to be auto merged. And it turns out that there's a bug when a test fails.

```
+ fail_test 'Elasticsearch pilot did not update the document count'
+ FAILURE_COUNT=1
+ echo 'TEST FAILURE: Elasticsearch pilot did not update the document count'
+ '[' 1 -gt 0 ']'
+ fail_and_exit test-elasticsearchcluster-1513251127-21582
+ local namespace=test-elasticsearchcluster-1513251127-21582
+ test_logged_errors test-elasticsearchcluster-1513251127-21582
/home/richard/go/src/github.com/jetstack/navigator/hack/libe2e.sh: line 134: test_logged_errors: command not found

```

wallrj:
I've done a quick fix for that in https://github.com/jetstack/navigator/pull/176

PR #183, opened 2018-01-08
Refactor genericpilot into modular subpackages

munnerz:

This refactors genericpilot into modular subpackages. I've done this so that we can more easily deviate from the 'GenericPilot' implementation, as I'm thinking it could become to restrictive.

This also tidies up the genericpilot package in general and separates out responsibilities more clearly.

I'm going to be adding a leaderelection package soon and attempt to integrate it with the Elasticsearch pilot. There may be further changes to genericpilot (i.e. shifting of responsibilities), but that can be in follow up PRs.

The API surface for GenericPilot is mostly the same, so the cassandra pilot should not have issues rebasing on this.

**Release note**:
```release-note
NONE
```


munnerz:
This is now working a lot more reliably than before.

There's currently a bug upon SIGTERM being sent that results in immediate subprocess exit (calling process.Wait multiple times). I'm going to refactor the processmanager package significantly in a follow up PR to have a signature closer to:

```go
type ProcessManager interface {
    // Start will start the underlying subprocess
    Start() error
    // Stop will stop the underlying subprocess and wait for it to exit
    Stop() error
    // Wait will return a channel that closes upon the subprocess exiting.
    // If the subprocess has not been started yet, the returned chan will
    // not close until the subprocess has been started and then stopped.
    Wait() <-chan struct{}
}
```

We can look at adding more things in future if needs be, but I'd rather keep the process manager interface as simple as possible for now.

/test e2e

pkg/pilot/genericpilot/genericpilot.go
```diff
@@ -83,28 +43,44 @@ func (g *GenericPilot) Run() error {
...
 
-	var wg sync.WaitGroup
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		wait.Until(g.worker, time.Second, g.Options.StopCh)
-	}()
+	ctrlStopCh := make(chan struct{})
+	defer close(ctrlStopCh)
+	go g.controller.Run(ctrlStopCh)
```
wallrj: 'g.controller.Run' returns an error, which needs dealing with.

wallrj: And there probably needs to be something to wait for that function to return.

pkg/pilot/genericpilot/genericpilot.go
```diff
@@ -83,28 +43,44 @@ func (g *GenericPilot) Run() error {
...
+		for {
+			// wait until the proces is running before calling Wait
+			if g.process != nil && g.process.Running() {
+				break
+			}
+			time.Sleep(2)
+		}
+		// we must call Wait else process.ProcessState won't be populated with
+		// exit code/details of the process.
+		g.process.Wait()
```
wallrj: This also returns an error, which needs handling.

munnerz: This is now resolved as part of the later commit which improves processmanager

pkg/pilot/genericpilot/genericpilot.go
munnerz: Done

munnerz:
New output of errcheck:

```
➜  navigator git:(refactor-genericpilot) ✗ errcheck ./pkg/pilot/...
pkg/pilot/elasticsearch/v5/write_config.go:63:16:	defer in.Close()
pkg/pilot/genericpilot/healthz.go:10:11:	}).Listen()
pkg/pilot/genericpilot/healthz.go:16:11:	}).Listen()
pkg/pilot/genericpilot/probe/listen.go:20:12:	w.Write([]byte(err.Error()))
```

I'm going to be refactoring/rejigging healthz endpoints (which covers the last 3), and the top one is as part of a defer (plus outside of the 'scope' of this PR imo)

munnerz:
I restarted the build infra workers

/retest

PR #186, opened 2018-01-09
Perform leaderelection in GenericPilot

### LINKED ISSUE #97: Providing leader election mechanisms to Pilots
Some databases require cluster-wide options to be set by a single node in the cluster. For example, Elasticsearch has an 'exclude shards' field that can be set via the Elasticsearch API, that tells the cluster to evict shards from the nodes listed.

Updates to this field need co-ordinating, and so having a single cluster-aware component that watches *all* Pilots is desired. I'm thinking we can embed this logic into the actual Pilot itself to save deploying additional components. This gives us HA by default too (up to the degree that the database itself is HA).

Rough requirements:

- Add a ConfigMap or Endpoints like type to navigator-apiserver so leader election can be performed irrespective of the k8s API server
- Import leader election code into the `genericpilot`
- Provide a `LeaderElectedSyncFunc` capability in the `genericpilot`, that will be called upon every update to every Pilot within the cluster the Pilot is a member of

It's then up to the actual Pilot implementation to do something sensible here. Additionally, the leader elected sync function would be 'allowed' to update the status block of any of the Pilot resources it processes.

Interested to hear your thoughts @wallrj - is this sort of thing required to handle scale down of Cassandra clusters, or say triggering backups or any other administrative action that may need to be performed?

/kind feature

munnerz:

Adds a leaderelection package and implements leader election in the GenericPilot.

Closes #97

**Release note**:
```release-note
Add support for leader election in pilots
```


munnerz:
@wallrj this is ready for review now.

I think the best way for us to write tests for this is probably with integration tests for GenericPilot itself. The leaderelection package doesn't really need much testing as it is largely just a convenient proxy to the upstream kubernetes package.

We can then provide our own SyncFunc and LeaderElectedSyncFunc, run two instances of GenericPilot and ensure that only one of them receives calls to LeaderElectedSyncFunc.

pkg/pilot/genericpilot/genericpilot.go
```diff
@@ -62,6 +64,12 @@ func (g *GenericPilot) Run() error {
 		} else {
 			glog.V(4).Infof("Control loop unexpectedly exited")
 		}
+	case err = <-g.runElector(ctrlStopCh):
+		if err != nil {
+			glog.V(4).Infof("Leader elector failed with error: %s", err)
+		} else {
+			glog.V(4).Infof("Leader elector unexpectedly exited")
```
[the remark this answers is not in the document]
wallrj: So as soon as the the `elector` go routing exits (e.g. with an error) the `Run` method will exit and close the stop chanel.
But there doesn't seem to be anything here that waits fot the controller and the process waiter go routines to finish.
Shouldn't there be a waitgroup or something here?

pkg/pilot/genericpilot/genericpilot.go
```diff
@@ -95,8 +103,20 @@ func (g *GenericPilot) runController(stopCh <-chan struct{}) <-chan error {
...
 	go func() {
 		defer close(out)
-		res := g.controller.Run(stopCh)
-		out <- res
+		out <- g.controller.Run(stopCh)
+	}()
+	return out
+}
+
+func (g *GenericPilot) runElector(stopCh <-chan struct{}) <-chan error {
```
wallrj: The stopCh isn't used.
Pass it to `elector.Run` ?

pkg/pilot/genericpilot/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+var _ Interface = &Elector{}
+
+func (e *Elector) Run() error {
+	// Identity used to distinguish between multiple controller manager instances
+	id, err := os.Hostname()
+	if err != nil {
+		return fmt.Errorf("error getting hostname: %s", err)
+	}
```
wallrj: Hostname is unique to the Pod running the pilot process, right?
What will it be? The Pod name?
In which case, it might be better to use the supplied `--pilot-name`, especially when we get round to writing tests for this.

pkg/pilot/elasticsearch/v5/controller.go
```diff
@@ -39,3 +38,8 @@ func (p *Pilot) syncFunc(pilot *v1alpha1.Pilot) error {
 	}
 	return nil
 }
+
+func (p *Pilot) leaderElectedSyncFunc(pilot *v1alpha1.Pilot) error {
+	glog.V(4).Infof("ElasticsearchController: leader elected sync of pilot %q", pilot.Name)
+	return nil
+}
```
wallrj: What operations might be performed here?
https://github.com/jetstack/navigator/issues/97 talks about `excludeShards`, but that seems like something that would be performed by the Pilot that had been selected to leave the cluster  by the ES controller.

munnerz: So in the ES pilot case, excludeShards will be handled here. I've got some more work locally to actually do this. This needs to be leader elected as it requires updating the ES API with the new list of nodes that should be excluded. We *could* do it in each individual Pilot and just hope there's no races, but this way is far more concrete.

pkg/pilot/genericpilot/genericpilot.go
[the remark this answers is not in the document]
munnerz: So we don't explicitly wait for each goroutine to exit right now - instead we rely on the call to `stop` to perform any required shutdown operations (which may or may not include stopping the controller). The controller itself *does* get stopped, but not until after the call to `stop` has returned (`defer close(ctrlStopCh)` a few lines above causes this).

We can change this to explicitly wait for the controller to finish 'draining' it's workers, and perhaps we should. As we don't *currently* rely on that though I've left it out for now in an attempt to not further block up process exit. I can foresee us needing to change this like you say in future though, as the controller is responsible for reporting Pilot status etc.

pkg/pilot/genericpilot/genericpilot.go
munnerz: `elector.Run` does not accept a stopCh (and is an upstream package) - I included the chan as an arg here for consistency with the other `run*` methods, and in case we do/can use it in future. Can be removed though as like you say, it's not used.

pkg/pilot/genericpilot/leaderelection/leaderelection.go
munnerz: I'm not 100% sure on this one. Pilot name *does* make more sense, however if there is a Deployment managing our Pods, and the update strategy is set to rollingUpdate, there are instances where two pods can be running at once (and both would have the same pilot name). Typically hostname is used when using the leader election packages as it should be unique (hence why we use Hostname when performing this same leader election in navigator-controller)

pkg/controllers/elasticsearch/nodepool/resources.go
```diff
@@ -177,6 +178,11 @@ func elasticsearchPodTemplateSpec(controllerName string, c *v1alpha1.Elasticsear
 							Name:  "PLUGINS",
 							Value: plugins,
 						},
+						{
+							Name: "LEADER_ELECTION_CONFIG_MAP",
+							// TODO: trim the length of this string
+							Value: fmt.Sprintf("%s-leaderelection", c.Name),
```
munnerz: Note to self: this config map name is not unique enough

Issue #106, opened 2017-11-08
ElasticSearch TestCreatesRoleBinding sometimes fails


wallrj:
```
...
go test $(go list ./... | grep -v '/vendor/')
?   	github.com/jetstack/navigator/cmd/apiserver	[no test files]
?   	github.com/jetstack/navigator/cmd/controller	[no test files]
?   	github.com/jetstack/navigator/cmd/controller/app	[no test files]
?   	github.com/jetstack/navigator/cmd/controller/app/options	[no test files]
?   	github.com/jetstack/navigator/cmd/pilot-elasticsearch	[no test files]
?   	github.com/jetstack/navigator/pkg/apis/navigator	[no test files]
?   	github.com/jetstack/navigator/pkg/apis/navigator/install	[no test files]
?   	github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1	[no test files]
?   	github.com/jetstack/navigator/pkg/apiserver	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/internalversion	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/internalversion/fake	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/internalversion/scheme	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/internalversion/typed/navigator/internalversion	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/internalversion/typed/navigator/internalversion/fake	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/versioned	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/versioned/fake	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/versioned/scheme	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/versioned/typed/navigator/v1alpha1	[no test files]
?   	github.com/jetstack/navigator/pkg/client/clientset/versioned/typed/navigator/v1alpha1/fake	[no test files]
?   	github.com/jetstack/navigator/pkg/client/informers/externalversions	[no test files]
?   	github.com/jetstack/navigator/pkg/client/informers/externalversions/internalinterfaces	[no test files]
?   	github.com/jetstack/navigator/pkg/client/informers/externalversions/navigator	[no test files]
?   	github.com/jetstack/navigator/pkg/client/informers/externalversions/navigator/v1alpha1	[no test files]
[...]

PR #187, opened 2018-01-09
Use label selector when listing pods in es controller

munnerz:

Use a label selector in the Elasticsearch navigator-controller implementations when listing pods to create Pilots for.

This also simplifies the Pilot creation code and decouples it from node pools.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controllers/elasticsearch/nodepool/controller.go
+++ b/pkg/controllers/elasticsearch/nodepool/controller.go
@@ -8,7 +8,6 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
 	appslisters "k8s.io/client-go/listers/apps/v1beta1"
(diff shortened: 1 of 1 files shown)


pkg/controllers/elasticsearch/nodepool/controller.go
```diff
@@ -172,40 +172,34 @@ func (e *statefulControl) existingStatefulSet(c *v1alpha1.ElasticsearchCluster,
...
+		return err
+	}
+
+	allPods, err := e.podLister.Pods(c.Namespace).List(selector)
 	if err != nil {
 		return err
 	}
+
 	for _, pod := range allPods {
 		isMember, err := util.PodControlledByCluster(c, pod, e.statefulSetLister)
```
wallrj: `isMember` is now unused and causing compile errors.

pkg/controllers/elasticsearch/nodepool/controller.go
```diff
@@ -172,40 +172,34 @@ func (e *statefulControl) existingStatefulSet(c *v1alpha1.ElasticsearchCluster,
...
 
 		if err != nil {
 			return fmt.Errorf("error checking if pod is controller by elasticsearch cluster: %s", err.Error())
 		}
 
-		if isMember {
```
wallrj: Isn't it important to *also* check that the pod is owned by the esCluster resource?
I thought this check was to ensure against deleting pods that, *somehow*, had been created independently, with the matching cluster label.
And if not, we've got code that looks up the owner of the owning statefulset; can that now be removed.

munnerz: Yep - this shouldn't have been removed. I've added that check back in.

pkg/controllers/elasticsearch/nodepool/controller.go
munnerz: Ack - accidentally removed too much code!

munnerz:
Thanks for taking a look @wallrj. I've updated to reintroduce the ownership check as it is still necessary. Regarding two ownerReferences - it would be nice, but also quite complex to do via a StatefulSet pod template (and perhaps not possible?). For now I think it's easiest to keep the behaviour the same.

munnerz:
Test flake caused by #106. I've opened #188 which should resolve it.

PR #189, opened 2018-01-10
WIP: Add a Cassandra Demo

wallrj:
**Release note**:
```release-note
NONE
```


munnerz:
Can this be closed, or otherwise be turned into a quick start doc and published in our docs? Looks like useful info if we don't already have it, but I think should be more visible than in `demo.sh` 😄

PR #195, opened 2018-01-11
Report ElasticsearchCluster health status to API

### LINKED ISSUE #191: Surface overall cluster status to the ElasticsearchCluster resource
We need a way to surface Elasticsearch cluster level information to the ElasticsearchCluster resource. Without teaching navigator-controller how to interact with Elasticsearch itself, which is a pattern we've avoided thus far, the `elasticsearchcluster.status.clusterHealth` (or equivalent) field needs to be updated appropriately.

The 'master' Pilot could update this field, but we'd then need to also ensure `navigator-controller` and the master pilot don't flip-flop performing updates to the status block of the resource.

I'm fairly sure a similar pattern exists between kube-controller-manager and kubelet, where both update the `node.status` block. This could perhaps be done by namespacing all of the fields the pilot is responsible for under a single yaml key (e.g. `elasticsearchcluster.status.cluster.health`.

This information could then also be used to influence the scale-up/scale-down decision or any cluster 'shape' change decisions.

/kind feature
/area elasticsearch

/cc @wallrj

munnerz:

Adds support for recording the health status of the Elasticsearch cluster back to the API.

This also introduces a 10s collection delay to the collection of node metrics for Elasticsearch.

**Which issue this PR fixes**:
fixes #191 

**Release note**:
```release-note
Record overall Elasticsearch cluster health as status on the ElasticsearchCluster resource
```

### DIFF
--- a/pkg/controllers/util.go
+++ b/pkg/controllers/util.go
@@ -1,8 +1,7 @@
 package controllers
 
 import (
-	"reflect"
-
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
@@ -32,10 +31,11 @@ func (q *QueuingEventHandler) OnAdd(obj interface{}) {
 }
 
 func (q *QueuingEventHandler) OnUpdate(old, new interface{}) {
-	if reflect.DeepEqual(old, new) {
-		return
+	oldObj := old.(metav1.Object)
+	newObj := new.(metav1.Object)
+	if oldObj.GetResourceVersion() != newObj.GetResourceVersion() {
+		q.Enqueue(new)
 	}
-	q.Enqueue(new)
 }
 
 func (q *QueuingEventHandler) OnDelete(obj interface{}) {
@@ -62,10 +62,11 @@ func (b *BlockingEventHandler) OnAdd(obj interface{}) {
 }
 
 func (b *BlockingEventHandler) OnUpdate(old, new interface{}) {
-	if reflect.DeepEqual(old, new) {
-		return
+	oldObj := old.(metav1.Object)
+	newObj := new.(metav1.Object)
+	if oldObj.GetResourceVersion() != newObj.GetResourceVersion() {
+		b.WorkFunc(new)
 	}
-	b.WorkFunc(new)
 }
 
 func (b *BlockingEventHandler) OnDelete(obj interface{}) {
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -145,7 +145,7 @@ function test_elasticsearchcluster() {
     then
         fail_test "Elasticsearch pilot did not update the document count"
     fi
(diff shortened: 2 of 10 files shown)


pkg/controllers/util.go
```diff
@@ -32,10 +31,11 @@ func (q *QueuingEventHandler) OnAdd(obj interface{}) {
 }
 
 func (q *QueuingEventHandler) OnUpdate(old, new interface{}) {
-	if reflect.DeepEqual(old, new) {
-		return
+	oldObj := old.(metav1.Object)
+	newObj := new.(metav1.Object)
```
wallrj: Why not add `metav1.Object` to the function signature instead of `interface{}` ?

pkg/controllers/util.go
```diff
@@ -62,10 +62,11 @@ func (b *BlockingEventHandler) OnAdd(obj interface{}) {
 }
 
 func (b *BlockingEventHandler) OnUpdate(old, new interface{}) {
-	if reflect.DeepEqual(old, new) {
-		return
+	oldObj := old.(metav1.Object)
+	newObj := new.(metav1.Object)
```
wallrj: Same here.

pkg/controllers/util.go
munnerz: We use this as an AddFunc, which has the signature `func(interface{})`. Go cannot realise that `func(metav1.Object)` matches `func(interface{})`.

munnerz:
1) 👍
2) The 'scheduledWorkQueue' takes care of this. It ensures that 'member pilots' (i.e. pilots that are a member of the same cluster) will be synced every ~10s regardless of resourceVersion etc.
3) Same as above
4) Sam as above
5) It would be great if it did have this (a watch mechanism). I think there is something like this in the client library we are using, but under the hood it's just polling afaik.
6) `How do we justify adding this health status field, while at the same time reducing the API surface area in #205 ?` <- not sure what you mean here. We're looking to *simplify*, not reduce the API surface. By no means does that mean we should halt future development, more just take count of what we have so far.
7) It *will* be used by Navigator to make decisions around scale up, but the information can also be surfaced to users (e.g. through `kubectl get` in future):

```
$ kc get esc
NAME               READY     HEALTH   MASTERS   INGEST   DATA
test-cluster       5/5       Green    3         2        2
```

8) Yes, it would be used to feed into things such as the update operator, or deciding whether to scale up/down. The delay here is particularly difficult, however I think we're *always* going to have to suffer some delay unless we fully block/shut down the database. Instead of aiming to get a perfect in time snapshot of the system, I'm *trying* to go down the route of *expecting* things to be out of date, but 'tolerating' a certain upper bound of late data (e.g. I will accept this as the truth *if* it was reported within the last 5s).

9) Regarding *how long* something has been green/yellow/red for, absolutely agree we need this information and it will be coming in a follow up. This is effectively the heartbeat. I'll be adding this further down the line for ESC :)

Thanks for the review Richard! Hopefully that clears up a few Qs. I'll get an e2e test together today.

munnerz:
argh. failed on some stupid networking flake...

/test e2e v1.9

Issue #194, opened 2018-01-11
Cleaning up v1alpha1 API in preparation for a 'stable' cut


munnerz:
We want to cut a 'stable' v1alpha1 API that can be depended upon, and begin working in the v1alpha2 API.

Before we do this, in the brief remaining window where we can perform breaking API changes, we should make our 'last' API changes to v1alpha1.

I'd like to call a review on all of our resources:

- Pilot

```go
type PilotSpec struct {
	Elasticsearch *PilotElasticsearchSpec `json:"elasticsearch"`
}

type PilotPhase string

const (
	// PreStart occurs before the Pilot's subprocess has been started.
	PilotPhasePreStart PilotPhase = "PreStart"
	// PostStart occurs immediately after the Pilot's subprocess has been
	// started.
	PilotPhasePostStart PilotPhase = "PostStart"
	// PreStop occurs just before the Pilot's subprocess is sent a graceful
	// termination signal. These hooks will block termination of the process.
	PilotPhasePreStop PilotPhase = "PreStop"
	// PostStop occurs after the Pilot's has stopped. These can be used to
	// clean up, or whatever other action that may need to be performed.
	PilotPhasePostStop PilotPhase = "PostStop"
)

type PilotElasticsearchSpec struct {
}

type PilotStatus struct {
	LastCompletedPhase PilotPhase       `json:"lastCompletedPhase"`
	Conditions         []PilotCondition `json:"conditions"`
	// Contains status information specific to Elasticsearch Pilots
	Elasticsearch *ElasticsearchPilotStatus `json:"elasticsearch,omitempty"`
}

type ElasticsearchPilotStatus struct {
	// Documents is the current number of documents on this node. nil indicates
	// an unknown number of documents, whereas 0 indicates that the node is
	// empty
	Documents *int64 `json:"documents,omitempty"`
}

// PilotCondition contains condition information for a Pilot.
type PilotCondition struct {
	// Type of the condition, currently ('Ready').
	Type PilotConditionType `json:"type"`

	// Status of the condition, one of ('True', 'False', 'Unknown').
	Status ConditionStatus `json:"status"`

	// LastTransitionTime is the timestamp corresponding to the last status
	// change of this condition.
	LastTransitionTime metav1.Time `json:"lastTransitionTime"`

	// Reason is a brief machine readable explanation for the condition's last
	// transition.
	Reason string `json:"reason"`

	// Message is a human readable description of the details of the last
	// transition, complementing reason.
[...]

PR #196, opened 2018-01-11
elasticsearch: Replace 'config' field with 'minimumMasters'

munnerz:

This PR alters the API surface for ElasticsearchCluster. It replaces the config field on node pools with a cluster wide minimum masters field.

The official Elastic recommendation is to modify the majority of cluster settings through the cluster API, and without a specific request to have a method to specify *static* config, I think we're best to remove this field for now.

Related to #194 

**Release note**:
```release-note
Remove 'config' field from node pools and replace with cluster-wide 'minimumMasters' field
```


pkg/apis/navigator/validation/elasticsearch_test.go
```diff
@@ -312,52 +332,58 @@ func TestValidateElasticsearchPersistence(t *testing.T) {
...
+			Pilot:          validSpecPilotImage,
+			MinimumMasters: 2,
+		},
+		"minimum masters set to low": {
+			Plugins:        validSpecPluginsList,
+			NodePools:      []navigator.ElasticsearchClusterNodePool{newValidNodePool("test", 3, navigator.ElasticsearchRoleMaster)},
+			Pilot:          validSpecPilotImage,
+			Image:          validSpecESImage,
+			MinimumMasters: 1,
+		},
```
wallrj: I wonder if there's ever a case where you want to run with fewer than quorum masters.
I mean if we're allowing a cluster with a single master why not allow a cluster, maybe for testing, with 3 masters but where only one is required to be running?

pkg/controllers/elasticsearch/configmap/resources.go
```diff
@@ -16,6 +34,49 @@ func esConfigConfigMap(c *v1alpha1.ElasticsearchCluster, np *v1alpha1.Elasticsea
...
+	minimumMasters := c.Spec.MinimumMasters
+	if minimumMasters == 0 {
+		// auto-manage minimum master count
+		totalMasters := countMasterReplicas(c.Spec.NodePools)
+		minimumMasters = calculateQuorom(totalMasters)
+	}
+	return fmt.Sprintf(configTemplate, minimumMasters)
+}
+
+// TODO: move these functions into a shared package
```
wallrj: Yeah, I think we should do that before merging this.

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -122,6 +122,12 @@ type ElasticsearchClusterSpec struct {
 	// Sysctl can be used to specify a list of sysctl values to set on start-up
 	// This can be used to set for example the vm.max_map_count parameter.
 	Sysctl []string `json:"sysctl"`
+
+	// The minimum number of masters required to form a quorom in the cluster.
+	// If omitted, this will be set to a quorom of the master nodes in the
+	// cluster. If set, the value *must* be greater than or equal to a quorom
+	// of master nodes.
```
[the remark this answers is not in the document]
munnerz: No - the field `minimumMasters` must be greater than or equal to a quorum of master nodes. If it were less than, a split brain could occur.

> And if so, how can you have a quorum greater than the number of masters in the cluster?

Exactly - if a user were to specify a `minimumMasters` parameter that is greater than the total number of masters in the cluster, a leader would never be elected.

pkg/apis/navigator/validation/elasticsearch.go
```diff
@@ -96,9 +98,53 @@ func ValidateElasticsearchClusterSpec(spec *navigator.ElasticsearchClusterSpec,
...
+	case numMasters == 0:
+		allErrs = append(allErrs, field.Invalid(npPath, numMasters, "must be at least one master node"))
+	case spec.MinimumMasters == 0:
+		// do nothing, navigator-controller will automatically calculate &
+		// manage the minimumMasters required for the cluster
+	case spec.MinimumMasters < quorom:
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("minimumMasters"), spec.MinimumMasters, fmt.Sprintf("must be a minimum of %d to avoid a split brain scenario", quorom)))
+	case spec.MinimumMasters > quorom:
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("minimumMasters"), spec.MinimumMasters, fmt.Sprintf("cannot be greater than the total number of master nodes")))
+	}
```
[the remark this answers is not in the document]
munnerz: 👍

pkg/apis/navigator/validation/elasticsearch_test.go
munnerz: This should never be allowed with Elasticsearch - I've created clusters with 3 nodes and a minimum of 1 master, and immediately got a split brain (two nodes elected themselves leader)

pkg/controllers/elasticsearch/configmap/resources.go
munnerz: I've put something together for this, but it won't actually reduce duplication between validation and this package as they work on different types.

Validation is performed on internal types, and due to lack of generics, there's no way to share these functions (except for CalculateQuorum).

munnerz: I've pushed up my changes which include this. For now, I've kept the two functions that work on API types in validation, and also created `pkg/util/api` to house the v1alpha1 versions of these functions (which the `configmap` package now depends upon).

I did consider converting the types using the generated conversion functions, but this seems excessive for what is effectively a couple of for loops.

PR #198, opened 2018-01-11
Use release-1.9 kubernetes branches

munnerz:

Update to use the newest Kubernetes release branches.

**Release note**:
```release-note
Upgrade to release-1.9 Kubernetes client libraries
```


munnerz:
@wallrj what version of dep are you using? I am on 0.3.2 and there is no newer release available.

munnerz:
So the hash actually was incorrect. #211 fixes this.

PR #199, opened 2018-01-11
Add basic ElasticsearchCluster resource validation

munnerz:

Adds an initial level of validation for ElasticsearchCluster resources.

Example output:
```
➜ kc create -f hack/testdata/es-cluster-test.template.yaml
The ElasticsearchCluster "test" is invalid:
* spec.pilot.tag: Required value
* spec.pilot.repository: Required value
* spec.nodePools[0].roles[2]: Unsupported value: "car": supported values: data, ingest, master
* spec.nodePools[1].name: Duplicate value: "mixed"
```

**Release note**:
```release-note
Add resource validation for ElasticsearchCluster resources
```


pkg/apis/navigator/validation/elasticsearch.go
```diff
@@ -0,0 +1,87 @@
...
+	case navigator.ElasticsearchRoleMaster:
+	default:
+		el = append(el, field.NotSupported(fldPath, r, supportedESClusterRoles))
+	}
+	return el
+}
+
+func ValidatElasticsearchClusterNodePool(np *navigator.ElasticsearchClusterNodePool, fldPath *field.Path) field.ErrorList {
+	el := ValidateDNS1123Subdomain(np.Name, fldPath.Child("name"))
+	el = ValidateElasticsearchPersistence(&np.Persistence, fldPath.Child("persistence"))
```
wallrj: This clobbers any errors returned from `ValidateDNS1123Subdomain`.

pkg/apis/navigator/validation/elasticsearch.go
```diff
@@ -0,0 +1,87 @@
...
+	// struct
+	return el
+}
+
+func ValidateElasticsearchPersistence(cfg *navigator.ElasticsearchClusterPersistenceConfig, fldPath *field.Path) field.ErrorList {
+	el := field.ErrorList{}
+	if cfg.Enabled && len(cfg.Size) == 0 {
+		el = append(el, field.Required(fldPath.Child("size"), ""))
+	}
+	// TODO: validate size quantity
```
wallrj: Can't we declare `Size resource.Quantity` in types.go instead?

Then check here that `cfg.Size.GreaterThan(0)`

pkg/registry/navigator/escluster/strategy.go
```diff
@@ -97,14 +99,8 @@ type esClusterStatusStrategy struct {
...
-	newESCluster, ok := new.(*navigator.ElasticsearchCluster)
-	if !ok {
-		glog.Fatal("received a non-elasticsearchcluster object to update to")
-	}
-	oldESCluster, ok := old.(*navigator.ElasticsearchCluster)
-	if !ok {
-		glog.Fatal("received a non-elasticsearchcluster object to update from")
-	}
+	newESCluster := new.(*navigator.ElasticsearchCluster)
+	oldESCluster := old.(*navigator.ElasticsearchCluster)
```
wallrj: I guess it's safe to assume that it'll always be supplied with the right types? Or is this only because of the added validation in this branch?

munnerz: My reasoning here: previously we called Fatal (which calls os.Exit). Now we call panic. It was always going to bring everything down, we've just changed how now and tidied things up. FWIW, yes we should always be passed the correct type. Upstream just directly casts (like this change makes us do), so figured we should do similar :)

pkg/apis/navigator/validation/elasticsearch.go
munnerz: :+1: Ack

pkg/apis/navigator/validation/elasticsearch.go
munnerz: Done, and validation added.

pkg/apis/navigator/validation/elasticsearch.go
```diff
@@ -0,0 +1,107 @@
...
+	return el
+}
+
+func ValidateElasticsearchClusterSpec(spec *navigator.ElasticsearchClusterSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := ValidateImageSpec(&spec.Image.ImageSpec, fldPath.Child("image"))
+	allErrs = append(allErrs, ValidateImageSpec(&spec.Pilot.ImageSpec, fldPath.Child("pilot"))...)
+	npPath := fldPath.Child("nodePools")
+	allNames := sets.String{}
```
wallrj: Ooh, `sets`! That's useful.

pkg/apis/navigator/validation/elasticsearch.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func ValidateElasticsearchPersistence(cfg *navigator.ElasticsearchClusterPersistenceConfig, fldPath *field.Path) field.ErrorList {
+	el := field.ErrorList{}
+	if cfg.Enabled && cfg.Size.IsZero() {
+		el = append(el, field.Required(fldPath.Child("size"), ""))
+	}
+	if cfg.Size.Sign() == -1 {
+		el = append(el, field.Invalid(fldPath.Child("size"), cfg.Size, "must be greater than zero"))
+	}
```
wallrj: Perhaps instead check that the value is >0 and this could all be replaced with, `if cgf.Size.Sign() != 1`

The docs say:

```
// Sign returns 0 if the quantity is zero, -1 if the quantity is less than zero, or 1 if the
// quantity is greater than zero.

```

pkg/apis/navigator/validation/elasticsearch_test.go
```diff
@@ -0,0 +1,379 @@
...
+			Size: validSize,
+		},
+		{
+			StorageClass: "something",
+		},
+		{
+			Enabled: true,
+			Size:    validSize,
+		},
+	}
```
wallrj: Thanks for these tests.
I think that the size should be validated and be required to be > 0, regardless of whether persistence is enabled.

munnerz: So if persistence is enabled, size MUST be >0 (cannot be zero).
If persistence is *disabled*, we should accept the zero value for size, *or* an otherwise valid value (despite it being ignored) but we should NEVER allow a negative value. I'll update/ensure our tests cover these cases.

pkg/apis/navigator/validation/elasticsearch.go
munnerz: As in my other comment, there are a few cases to cover:

1) Persistence enabled, size is >0 (VALID)
3) Persistence enabled, size = 0 (INVALID)
2) Persistence disabled, size = 0 (VALID)
4) Size<0 - ALWAYS invalid

I've expanded our test suite to explicitly cover all the cases above, and they are still passing 😄

munnerz:
You can see the unit test that ensures that size is valid *regardless* whether Enabled=true here: https://github.com/jetstack/navigator/pull/199/files#diff-37f8430e4ef53a4d14e2372ffa292324R262

We *have* to allow the zero value *iff* enabled = false so that users can write:
```yaml
persistence:
  enabled: false
```

pkg/apis/navigator/validation/elasticsearch.go
wallrj: I still don't see the case for `3 Persistence disabled, size = 0 (VALID)`
Unless that's how we determine that persistence is not required.
In which case, would that be better expressed as `Persistence  *ElasticsearchClusterPersistenceConfig` and `Persistence = nil` ?

wallrj: Ah, I see you just explained below.

munnerz: So if I set:

```persistence:
  enabled: false```

I have implicitly also said the size=0. It would we weird/awkward to require users to write:

```persistence:
  enabled: false
  # size is ignored
  size: 1Gi
```

PR #201, opened 2018-01-12
Rename fsGroup to runAsUser and set RunAsUser on ES pods

munnerz:

This renames the 'fsGroup' field to 'runAsUser', and additionally sets the `runAsUser` fields on pods created by the controller. This fixes issues where there may be a mistmatch between the fsGroup specified and the actual user in the image.

This is required for 6.x support

**Release note**:
```release-note
Rename 'fsGroup' to 'runAsUser' and add support for Elasticsearch 6.x
```

/area elasticsearch
/kind feature

### DIFF
--- a/hack/testdata/es-cluster-test.template.yaml
+++ b/hack/testdata/es-cluster-test.template.yaml
@@ -19,7 +19,7 @@ spec:
     ## This sets the group of the persistent volume created for
     ## the data nodes. This must be the same as the user that elasticsearch
     ## runs as within the container.
-    fsGroup: 1000
+    runAsUser: 1000
 
   nodePools:
   - name: mixed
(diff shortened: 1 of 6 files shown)


munnerz:
ref. #194

/test e2e

hack/testdata/es-cluster-test.template.yaml
```diff
@@ -18,7 +18,7 @@ spec:
     ## This sets the group of the persistent volume created for
     ## the data nodes. This must be the same as the user that elasticsearch
     ## runs as within the container.
-    fsGroup: 1000
+    runAsUser: 1000
```
wallrj: Also need to update the example yaml in `docs/`

munnerz:
1) Yep, exactly
2) That's what setting `fsGroup` is for (for the attached PVs). It's the users responsibility to ensure other files (e.g. ES config within the container) is owned by the appropriate user.
3) It doesn't actually require any root privs 😄

munnerz:
Closing in favour of #214

PR #205, opened 2018-01-13
Add 'version' field to ESC. Auto-generate Elasticsearch image details

munnerz:

This PR adds a version field to an ElasticsearchCluster. It is used to auto-default the ElasticsearchImageSpec in its absence. If a user wants to override the image to be used and take manual control, they can specify their own fields as before.

The new version field should be required regardless whether the image is manually specified, as in future it can be used to help influence upgrade strategy. It will be up to the administrator to update the image and listed version in unison.

In future we should make the default image repository and uid be configurable.

Ensuring the spec.version field is set will be in a follow up after #199 is merged

**Release note**:
```release-note
Add a new spec.version field to ElasticsearchCluster and auto-default the image version and repository to an official upstream image.
```


pkg/controllers/elasticsearch/nodepool/resources.go
```diff
@@ -133,6 +133,18 @@ func elasticsearchPodTemplateSpec(controllerName string, c *v1alpha1.Elasticsear
 	plugins := strings.Join(c.Spec.Plugins, ",")
 	nodePoolLabels := util.NodePoolLabels(c, np.Name, np.Roles...)
 
+	esImage := c.Spec.Image
+	if esImage.FsGroup == 0 &&
+		esImage.ImageSpec.PullPolicy == "" &&
+		esImage.ImageSpec.Repository == "" &&
+		esImage.ImageSpec.Tag == "" {
```
wallrj: * Why check for FsGroup here?
* Could you instead compare esImage.ImageSpec to the zero value of ImageSpec?
* What if I supply FsGroup, but omit the ImageSpec? It looks like the pod will fail to start.

munnerz: 1) Changing the `version` attribute is expected use. This should cause navigator to begin an upgrade/downgrade/report an error that upgrade/downgrade isn't possible.

2) Nope - provide the `version` and everything else is handled for you. As soon as you specify either FsGroup, PullPolicy, Repository or Tag however, you've gone into manual mode. Manual mode *might* be removed at some point in the future (as discussed today)

3) As above, fsGroup is checked because if specified, you've switched to 'manual mode' (and so all the others are required too). Perhaps this will make a bit more sense once we get the validation PR (#199) merged.

4) Yep potentially we could just check the zero values - or as you suggested earlier, make this field a pointer.

5) Yep you are correct. FsGroup = manual mode. Remember users never see `imageSpec` as a field (as it's json tag is: `json:",inline"`)

pkg/controllers/elasticsearch/nodepool/util_test.go
```diff
@@ -0,0 +1,126 @@
...
+		{
+			name: "invalid (non semver) version specified and no manual image",
+			spec: &v1alpha1.ElasticsearchClusterSpec{
+				Version: "not.semver",
+			},
+			expectedError: true,
+		},
+		// We don't error here - if the specified version is non semver but the
+		// user has specified a manual image, validation should detect the
+		// invalid semver and fail.
```
wallrj: Shouldn't we always error with non-semver version?
Or does this comment mean that the non-semver  version will be caught and rejected by resource  validation in #199?

pkg/apis/navigator/types.go
```diff
@@ -89,10 +89,11 @@ type ElasticsearchClusterList struct {
 }
 
 type ElasticsearchClusterSpec struct {
+	Version   string
```
wallrj: I'd be inclined to make this type `*semver.Version` ?

pkg/controllers/elasticsearch/nodepool/util_test.go
munnerz: Precisely that - yep. It seems odd to return an error in this case. The function should be returning the image that should be used for the spec, and the image is explicitly defined in the manifest, so there's not really an error.

There *is* however an error with the overall state/spec of the cluster, and validation should be responsible for detecting this. I guess more gates isn't a bad thing, but I think the gate might need to be moved to some other function if we do add it. I'm working on validation now and will expand that PR to encompass new fields added here once merged.

pkg/apis/navigator/types.go
munnerz: I'm not sure we can just make this some other struct. The Golang JSON parser would fail because `string` is not a structure, which it would expect the field to be.

munnerz: That said, UnmarshalJSON etc. is defined so *maybe* it would. I'll take a look. Depending on external types in our apis can get pretty prickly though - if we update our dependency it could cause breaking API changes (and we don't 'control' the downstream code, which makes it harder to fix/manage)

munnerz:
@wallrj I've updated the PR to make the new Version field in our API into a `semver.Version`.

The unit tests obviously work with Golang structs (and pass), but they don't test the unmarshalling from a string. I've added a `version` parameter to the Elasticsearch Cluster test fixture for our e2e tests in order to verify this works correctly, and from what I can see it is passing happily.

Once #199 merges, I'll add further validation to this to stop invalid values being submitted to the API.

munnerz:
I've added additional validation to this now, which ensures that `version` is specified.

Unmarshaling will error if an invalid string is provided, but it doesn't error if *no* string is provided (it returns nil). In order to handle this case, I check for an empty `semver.Version` struct during validation and through a `Required` error in this case.

pkg/apis/navigator/validation/elasticsearch_test.go
```diff
@@ -360,37 +378,59 @@ func TestValidateElasticsearchClusterSpec(t *testing.T) {
...
-			Image:          validSpecESImage,
+			Image:          &validSpecESImage,
 			MinimumMasters: 5,
 		},
+		"missing elasticsearch version": {
+			Plugins:   validPluginsList,
+			NodePools: []navigator.ElasticsearchClusterNodePool{validNodePool, validNodePool},
+			Pilot:     validPilotImage,
+			Image:     validESImage,
+		},
```
wallrj: Version is missing from all the other error cases too. 
So it's not clear whether they're failing because of version or minimummaster

Issue #203, opened 2018-01-12
Pilot upgrade testing strategy


munnerz:
One of navigators core goals is maintaining a stable and sane API that allows for upgrading systems over time. This includes underlying the database/application, the colocated Pilots, as well as Navigator itself.

In order to provide this level of confidence that it is safe to upgrade, we need a robust testing strategy in place as early on as possible.

PR #206, opened 2018-01-13
Add defaults for ElasticsearchCluster pilot image

munnerz:

Adds a defaulting function for the Elasticsearch pilot image spec. This makes `pilot` an optional field.

We need to think about how this will update over time. Currently, it's up to the end users of the cluster to ensure the pilot version used is compatible with/within a reasonable range of navigator.

navigator-controller needs to support periodically updating the pilot image specified on Cassandra/ElasticsearchCluster resources, thus triggering the regular cluster upgrade procedure. i'm going to open a separate issue to track this as this is a larger design decision.

**Release note**:
```release-note
Make 'spec.pilot' an optional field on ElasticsearchClusters
```

### DIFF
--- a/pkg/apis/navigator/types.go
+++ b/pkg/apis/navigator/types.go
@@ -126,7 +126,7 @@ type ElasticsearchClusterPersistenceConfig struct {
 type ImageSpec struct {
 	Repository string
 	Tag        string
-	PullPolicy string
+	PullPolicy v1.PullPolicy
 }
 
 type ElasticsearchPilotImage struct {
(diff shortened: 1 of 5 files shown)


munnerz:
Yep we should use omitempty - could call.

General Pilot upgrade strategy is discussed in #203 - this may be tricky, and might result in us moving the field altogether. Perhaps we should 'fully manage' this (as in, don't have a field on the resource at all and instead 'just do it')..... 🙄

munnerz:
Closing this for now. We need to think about how we handle versioning Pilot, and how we roll out upgrades.

/close

Issue #216, opened 2018-01-23
Cassandra data should be persisted to disk


wallrj:
The Cassandra statefulset / pods should have a persistent disk.

See https://github.com/jetstack/navigator/blob/master/pkg/controllers/elasticsearch/nodepool/resources.go#L110

```
if !np.Persistence.Enabled {
	volumes = append(volumes, apiv1.Volume{
		Name: esDataVolumeName,
		VolumeSource: apiv1.VolumeSource{
			EmptyDir: &apiv1.EmptyDirVolumeSource{},
		},
	})
}
```

/kind feature

Issue #224, opened 2018-01-25
Potential data loss due to CASSANDRA_AUTO_BOOTSTRAP=false


wallrj:
We should investigate, test and comment on https://github.com/kubernetes/examples/issues/175 

> Setting auto_bootstrap to false when bootstrapping a new node exposes the cluster to huge inconsistencies. This is because all the other steps in the process are carried out but no data is streamed to the node. Hence, the node would be in the UN state without having any data for the token ranges it has been allocated! Furthermore, the new node without data will be serving reads and nodes that previously owned the tokens will no longer be serving reads. Effectively, the token ranges for that replica would be replaced with no data.

http://thelastpickle.com/blog/2017/05/23/auto-bootstrapping-part1.html

If it is an issue we need to handle this safely in Navigator.

/kind bug

wallrj:
Note that in #222  we've switched to different image which does not allow the bootstrap mode to be configured from an environment variable.
We need to investigate what the default setting is in the Docker maintained images and how to set it....assuming it's necessary to set it.

Issue #222, opened 2018-01-25
Support official Cassandra Docker images


wallrj:
We're currently using the Google maintained Cassandra image from the Cassandra + Kubernetes demo.

* https://kubernetes.io/docs/tutorials/stateful-application/cassandra/

We should also support other, official, Cassandra V3 images such as:
 * https://hub.docker.com/_/cassandra/

One key difference in that image is that it has a different entrypoint.

So the hard coded `/run.sh` in https://github.com/jetstack/navigator/blob/master/pkg/pilot/cassandra/v3/pilot.go#L56 needs to be configurable somehow....or auto detected...if that's possible.

/kind feature

munnerz:
👍 although I am not interested in supporting multiple different images at this point. Differences in entrypoint, config directory etc. will be a pain to handle well as that information needs expressing on the *Cluster resource, and then passing all the way down and through.

I'd rather us focus on other core parts, and for now have a fixed series of Docker images that can be used. As and when we have a requirement for custom images, we can focus on that. For now, scale down/upgrade etc etc are much higher priority.

munnerz:
FWIW, I'm not against us switching which Docker images we use for this if you would prefer to use the 'official' ones (side note: who makes these images?).

Specifically I don't think we should complicate our APIs right now with supporting fields to allow this to be completely customised by the user, as I fear a user could start using an image that does something differently, and then when it comes to designing our upgrade process or the like, we have to make assumptions that are now false. By controlling it for now, we will gain an understanding of what requirements we need to put on the images that are used, so when it comes to adding support for arbitrary custom images we will have a bullet pointed list of requirements of the images that are used.

kragniz:
I'd rather switch to using better supported images. Cassandra from kubernetes/examples doesn't seem well supported, with the last updates from May of last year (currently on 3.10, while 3.11 was released in june).

The cassandra image on dockerhub is supported by docker, inc (https://docs.docker.com/docker-hub/official_repos/). They should provide security updates and general maintenance.

PR #225, opened 2018-01-25
Persistent volumes for Cassandra data

wallrj:
* A new CassandraCluster peristence configuration section in the Cassandra API and the Helm chart.
* Including settings for Size and StorageClass per node pool
* Added volume claim template to the  Cassandra nodepool / statefulset
* E2E tests that create a key space, table and data and verify that the data is still available after pod deletion.
* Various improvements to the argument handling, error handling and return code handling in the E2E test helper library.

Fixes: #216

**Release note**:
```release-note
NONE
```


wallrj:
@munnerz This is ready for an initial review. 
 * I haven't added PVCs etc yet because I haven't worked out how to test it.
 * And I need to investigate why those E2E tests failed.
 * As far as I can tell, my tests are correct, so I suspect that Cassandra might be answering CQL queries before it has loaded all its data....as mentioned in #224

munnerz:
This all looks good to me and makes sense (given you've not done the PVC bits yet). Can you tidy up commits though?

You aren't actually persisting any data as part of this pull request, so I'm not sure why we'd expect any data to be persisted (hence the e2e tests failing).

What kind of tests are you looking to make for PVCs? All that is required is adding a PersistentVolumeClaimTemplate field to the StatefulSet that gets created. A simple unit test to ensure this template is set, and the name matches the name of the volume mount should suffice. During e2e tests it isn't possible to use PVCs properly as we are using minikube, which doesn't provide a dynamic volume provisioner.

wallrj:
The emptydir should survive as long as the pod, right?

So by killing the Cassandra process I expect it to restart with the data
that had been inserted.

And the tests pass on my local cluster and on the 1.9 tests.

So I think there's something else going on here.

munnerz:
Looking good - one thing to note, I'm not sure if we need to bother testing whether the PVCs are bound to the PVs you've manually created in e2e tests. Doing so is not a test of Navigator, but a test of the StatefulSet controller (as well as the PVC<->PV controller). Can you see any value beyond this that it brings? I'd be happy to just have a unit test that ensures the StatefulSet has the expected PVC template specified, as it's then up to Kubernetes itself to ensure that is bound and mounted.

wallrj:
So the failed tests show data being lost.

```
https://00e9e64bac29f2cf25eca53f369005d15e3318c7ccdce1527e-apidata.googleusercontent.com/download/storage/v1/b/jetstack-logs/o/pr-logs%2Fpull%2Fjetstack_navigator%2F225%2Fnavigator-e2e-v1-9%2F188%2Fartifacts%2Fminikube-logs.txt?qk=AD5uMEtdEBaVgTGYhbVEXhgisaIrkVr4SAmfxSoHaNhWf8X-U3_d8OnmFZCPor_6p1rohvQrpaPj64--8nbMXaumJVmWB-k0rVKYCHg8yh3kzLOpPz8j0Nc58ZPqKG2JjPSBekrMkarwlLC9Ghpei79GN1BI_Dn40Awp65TAHj3-gCz6m5hl70Jh7p9XBq-ZrygMiBoaO9RPzqow1Cw54XTVyLXbiZCGL0BxagCQF-GFXChS0NPjx3BEJWajVY2oOozmGj4Mz69XFPg9DNw1hKbG2czppK1arJephCkuBpNiTWSjEXKlXd2E-goYAjUMWz8S3TzBV2R-7Uru9cZxdl3Fm4UIZhcOUYbaRPFOimJg7wPNt5-jCAMk4djRBP_gfH2HWCwCM_sBefrAZspShmkk8ZBr8izQnVT4tICxCCmyJFZauZeckqsipHqDMR-WNUhmLJNrkugWT9IEeR0Yt80TOAs9PCiygbQ3_zG9hVKyPb2jSEtZWErK_1wJSl3fY2y84nmAtmyoR3326GtKBxuEqEuXWBCU5Td3D5DZB7Ja4RV0H3Ls83LhBz7JY4mMWJIPnvcGCnB_gNU6_bFrsP3WJ_iYAGyPOhU2oXpF0UVN09vTfWCiLT3LiKs_Glb7_jCsof_ebgdTpA8GNPOc8eNlciuWWsldCvp8tsbgapsnTg24mBBLwzXDdNp-zmt5DB15cDRE_wp_wy5qonZxqrippB_SFgdZoT_0PEn8oUkCeU9xZPAK_VXJr0_Wbo4YX1yQkJf7HI39ElQwB0mMXhbEiizt8gVLyQzRc694yZabPLfRJ_CAKw85z8HvhSIYlWp7gEBkmGO0LsZLoRK59VETodhpa1COHQ
```
* https://jetstack-build-infra.appspot.com/build/jetstack-logs/pr-logs/pull/jetstack_navigator/225/navigator-e2e-v1-9/188/

The startup log of the Cassandra process in the replacement Pod shows the keyspace and table being restored and the commit log being replayed.

```
INFO  14:08:44 Initializing space1.testtable1
INFO  14:08:44 Not submitting build tasks for views in keyspace space1 as storage service is not initialized
INFO  14:08:44 Initializing system_auth.resource_role_permissons_index
INFO  14:08:44 Initializing system_auth.role_members
INFO  14:08:44 Initializing system_auth.role_permissions
INFO  14:08:44 Initializing system_auth.roles
INFO  14:08:44 Not submitting build tasks for views in keyspace system_auth as storage service is not initialized
INFO  14:08:44 Initializing system_traces.events
INFO  14:08:44 Initializing system_traces.sessions
INFO  14:08:44 Not submitting build tasks for views in keyspace system_traces as storage service is not initialized
INFO  14:08:44 Completed loading (1 ms; 15 keys) KeyCache cache
INFO  14:08:44 Replaying /cassandra_data/commitlog/CommitLog-6-1517407569595.log, /cassandra_data/commitlog/CommitLog-6-1517407569596.log
INFO  14:08:46 Log replay complete, 26 replayed mutations
```

But the previous Cassandra process obviously wasn't shutdown gracefully and didn't flush its commit log to disk.

The reason, I think, is that the SIGTERM is being propagated by the Pilot to an intermediate bash script, which doesn't then propagate it to the Cassandra process.

One way to fix this  is to use an image whose entrypoint execs the Cassandra process, so that Cassandra becomes the immediate child of the Pilot.

The Docker maintained image does this #222

wallrj:
@munnerz Ready for another review now that I've rebased on top of the official Docker image and squashed all the commits.

Some answers to your previous comments:

> I'm not sure if we need to bother testing whether the PVCs are bound to the PVs you've manually created in e2e tests. Doing so is not a test of Navigator, but a test of the StatefulSet controller (as well as the PVC<->PV controller). 

Only reason I'm creating the PVs is because we don't have dynamic volume provisioning on our E2E test cluster yet. See https://github.com/jetstack/test-infra/issues/134

> Can you see any value beyond this that it brings? 

Yeah, I think it's valuable to demonstrate that Navigator runs Cassandra in such a way that data isn't lost, even when all cluster nodes are stopped.

As you can see in the comment above, the E2E tests revealed that the Cassandra database wasn't being shut down gracefully and therefore not flushing its commit log to disk...leading to data loss...so I think that illustrates the value.

> I'd be happy to just have a unit test that ensures the StatefulSet has the expected PVC template specified, as it's then up to Kubernetes itself to ensure that is bound and mounted.

I didn't add unit tests yet, but I can also do that here, if you like, or in a followup branch if that's ok with you. So far I haven't bother writing unit tests verifying the exact resources that Navigator controllers create. Instead just verifying that e.g. a statefulset is created and that only owned statefulsets are modified.

I need to consider the case where a user tries to change the size or the storage class.
For now I think we should look at ways to disallow that.
I

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -32,10 +32,26 @@ type CassandraClusterSpec struct {
 	CqlPort   int32                      `json:"cqlPort"`
 }
 
+// CassandraClusterPersistenceConfig contains persistent volume
+// configuration.
+type CassandraClusterPersistenceConfig struct {
```
munnerz: Can we rename `ElasticsearchClusterPersistenceConfig` to `PersistenceConfig` and re-use it between the two implementations?

This should also support the 'Enabled' field (like `ElasticsearchClusterPersistenceConfig` has) to allow users to deploy without PVs for testing purposes etc.

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -44,6 +47,26 @@ func StatefulSetForCluster(
...
+							apiv1.ReadWriteOnce,
+						},
+						Resources: apiv1.ResourceRequirements{
+							Requests: apiv1.ResourceList{
+								apiv1.ResourceStorage: np.Persistence.Size,
+							},
+						},
+					},
+				},
+			},
```
munnerz: We should only be setting the PVC template if persistence is enabled, if it's not, we should add an extra `spec.template.spec.volumes` which is an emptyDir with the name `cassDataVolumeName` so that the volumeMount references a valid volume still.

docs/quick-start/cassandra-cluster.yaml
```diff
@@ -9,6 +9,9 @@ spec:
   nodePools:
   - name: "RingNodes"
     replicas: 3
+    persistence:
+      size: 5Gi
```
munnerz: Missing `enabled: true`

wallrj:
Thanks for the reviews @munnerz 

I've addressed your comments and pushed again.

Let's see if the E2E tests pass.

PR #228, opened 2018-01-29
Refactor Elasticsearch controller to use a semi-state machine design & add ginkgo based e2e tests

### LINKED ISSUE #32: Switch e2e tests to use ginkgo
By writing our e2e tests in Go, we develop far more intricate tests. We should look to rewrite our existing e2e tests into the same format, and from there look at writing more tests.

### LINKED ISSUE #156: ElasticSearch E2E tests fail on non-minikube cluster
The ElasticSearch E2E tests rely on finding a jetstackexperimental/pilot-elasticsearch Docker image.

But on GCE I'm pushing the newly built images to e.g. `gcr.io/jetstack-sandbox/pilot-elasticsearch:build`

It'd be great if we could over ride the pilot image repo in the tests.

We could do that by adding an ElasticSearch helm chart with overridable image and tag.

### LINKED ISSUE #215: Design/plan finite state machine for ElasticsearchCluster management
@wallrj, @kragniz and I discussed the possibility of representing all the different state transitions when a user alters an ElasticsearchCluster resource as a finite state automata.

This would involve, at it's heart, a big 'switch style' statement (perhaps actually a series of if/else, depending on what's required). Each 'case' would represent a destination state, predicated with certain conditions (i.e. user wants to change version from 5.6.2->5.6.3, and cluster state is currently yellow, etc.).

This should make it easier to reason about the state of the world before we take action, making it easier to reason about what we're going to do as well as testing and proving it works.

For now, we'll need to handle the 'deadlock' case (e.g. where no progress can be made as the state of the world is 'unhandled' by logging messages to users.

munnerz:

This change refactors the Elasticsearch controller to use a state-machine 'style' of operation.

We introduce an 'Action' interface with a simple spec:

```go
type Action interface {
	Name() string
	// Execute should attempt to execute the action. If it is not possible to
	// apply the specified changes (e.g. due to the cluster not being in a
	// 'ready state', or some transient error) then an error will be returned
	// so the action can be requeued. This allows for non-blocking blocking of
	// actions, with retries. The workqueues default scheduling and rate limit
	// will thus handle fairness within Navigator, and handle backing off on
	// retries.
	Execute(state *State) error
}
```

The `cluster_control.go` contains the 'state machine' like logic. For now, this is implemented as a simple function that uses if/else branches to determine which 'Action' to return depending on the current state of the cluster.

Fixes #215, fixes #156, fixes #32 

I'm currently working on a general purpose unit testing fixture, `StateFixture` that can be used to mock the `State` structure. This should allow for easy testing of all of our Action implementations, between Elasticsearch and in future Cassandra.

Right now, this PR does *not* properly wait for document count to be zero during scale down, however it *does* manage a rolling upgrade across multiple statefulsets by careful management of the `partition` field on statefulset.spec.updateStrategy.rollingUpdate`.

Things still to do:
- managed scale down
- support changing node pool resources
- support changing node pool roles
- support changing sysctls
- support changing pilot image
- support changing securityContext

**Release note**:
```release-note
Introduce a managed node pool upgrade process
```

/cc @mattbates @simonswine @wallrj


munnerz:
@wallrj this PR has grown large, so I'm not going to continue any more work here. Our e2e tests are passing and stand up an ElasticsearchCluster (both single and multi-node), plus wait for the cluster to transition into a Yellow/Green state respectively. I'll be following up with more PRs to address the remaining items:

* managed scale down
* support changing node pool resources
* support changing node pool roles
* support changing sysctls
* support changing pilot image
* support changing securityContext

I'll also be expanding the test suite further, to:

* test performing an upgrade
* test writing data to one node and reading from another
* test reading & writing data whilst the cluster is in a degraded state
* test updating resources/other fields in the cluster

This PR also introduces two new test frameworks:

1) the StateFixture, which is useful for unit/integration testing `controller.Actions`. I've also added e2e tests for the upgrade and scale actions, that ensure the controller behaves as expected under various 'initial states'

2) the ginkgo e2e test framework. This is based off the kubernetes/kubernetes test framework.

Once merged, we can start to migrate the remaining cassandra e2e tests to this new format too. It appears to be running stable, and cleans up after itself nice and quickly 😄 

Please reach out if you need a hand with review/if there's anything I can do to help. Big PRs are big 😬

munnerz:
Closing in favour of #239, #240, #241, #242

PR #229, opened 2018-01-31
Use the Docker maintained Cassandra 3 image

wallrj:
* Use the Docker maintained Cassandra 3 image.
* Use TCP connect based readiness and liveness probe because this image doesn't contain a nodetool based `/ready-probe.sh` script.
* These TCP connection checks will succeed even though the process is stopped; so I now simulate a node failure using `nodetool decommission` which makes cassandra stop listening on its CQL port.
* I can improve on this when #170 is merged. 

Fixes: #222 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/contrib/charts/cassandra/templates/cassandracluster.yaml
+++ b/contrib/charts/cassandra/templates/cassandracluster.yaml
@@ -5,8 +5,8 @@ metadata:
   labels:
     app: {{ template "name"  . }}
     chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
 spec:
   cqlPort: {{ .Values.cqlPort }}
   sysctls:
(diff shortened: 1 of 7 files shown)


contrib/charts/cassandra/templates/cassandracluster.yaml
```diff
@@ -15,10 +15,10 @@ spec:
   - name: ringnodes
     replicas: {{ .Values.replicaCount }}
   image:
-    repository: {{ .Values.image.repository }}
-    tag: {{ .Values.image.tag }}
-    pullPolicy: {{ .Values.image.pullPolicy }}
+    repository: {{ .Values.image.repository | quote }}
+    tag: {{ .Values.image.tag | quote }}
```
wallrj: I got helm lint errors when I used an unquoted integer here.

https://github.com/kubernetes/helm/blob/master/docs/charts_tips_and_tricks.md#quote-strings-dont-quote-integers suggests quoting all strings, so I decided to do that.

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -195,10 +179,6 @@ func StatefulSetForCluster(
 									Name:  "CASSANDRA_RACK",
 									Value: "Rack1-K8Demo",
 								},
-								{
-									Name:  "CASSANDRA_AUTO_BOOTSTRAP",
-									Value: "false",
-								},
```
wallrj: The Docker image doesn't pay attention to this environment variable so I removed it.

We'll see if it causes any dataloss problems when I rebase #225

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -53,9 +53,7 @@ func (p *Pilot) Hooks() *hook.Hooks {
 }
 
 func (p *Pilot) CmdFunc(pilot *v1alpha1.Pilot) (*exec.Cmd, error) {
-	// The /run.sh script is unique to gcr.io/google-samples/cassandra:v12.
-	// TODO: Add support for other Cassandra images with different entry points.
-	cmd := exec.Command("/run.sh")
+	cmd := exec.Command("/docker-entrypoint.sh")
```
munnerz: I guess we’ll still have issues handling SIGTERM with this? Unless they have properly set up signal handling in this script.

PR #238, opened 2018-02-06
Use k8s sets package in install_plugins.go

munnerz:

Switch to use sets in install_plugins

**Release note**:
```release-note
NONE
```


munnerz:
Grr.. not sure why we keep seeing timeouts when deleting the test namespace.

/retest

PR #239, opened 2018-02-06
Switch replicas fields to use int32 instead of int64

munnerz:

Given `replicas` fields on the workloads API objects are all int32, this is in keeping with that and saves the loss of precision.

**Release note**:
```release-note
NONE
```


munnerz:
Argh this time it failed because the `kubectl run` to test a cassandra node failed to return stdout. I guess this was just a temporary network blip.

/retest

Issue #44, opened 2017-10-12
Support shard allocation awareness


munnerz:
Elasticsearch can have parameters set for shard allocation awareness, so that it can distribute shards according to various policies (e.g. across racks, AZs etc.)

We should support this through the ElasticsearchCluster resource. Ideally, we wouldn't introduce the concept of racks, but instead allow a user to be flexible through the use of labels.

https://www.elastic.co/guide/en/elasticsearch/reference/current/allocation-awareness.html

munnerz:
I've started mocking up how our API could look for this - so far I've come up with:

```yaml
apiVersion: navigator.jetstack.io/v1alpha1
kind: ElasticsearchCluster
metadata:
  name: test
spec:
  minimumMasters: 1
  sysctl:
  - vm.max_map_count=262144

  pilot:
    repository:
    tag:
    pullPolicy:

  image:
    repository: docker.elastic.co/elasticsearch/elasticsearch
    tag: 5.6.2
    pullPolicy: IfNotPresent
    ## This sets the group of the persistent volume created for
    ## the data nodes. This must be the same as the user that elasticsearch
    ## runs as within the container.
    runAsUser: 1000

  # == new field ==
  allocationAwareness:
    attributes:
    - zone
    - region
  # == end new field ==
  nodePools:
  - name: mixed
    replicas: 3

    roles:
    - master
    - ingest
    - data

    resources:
      requests:
        cpu: '500m'
        memory: 2Gi

    persistence:
      enabled: false

    # == new field ==
    attributes:
    - name: "zone"
      valueFrom:
        nodeLabel: "kubernetes.io/zone"
    - name: "region"
      valueFrom:
        nodeLabels: "kubernetes.io/region"
    - name: "use"
      value: "mixed"
    # == end new field ==
```

/cc @simonswine @wallrj @cehoffman wdyt of this? I've purposefully namespaces `allocationAwareness.attributes` so that we can support the `force.zone` fields of the ES API in future.

I think the structure for `nodeLabel` (under `attributes`) might need re-jigging to be a struct in case we need to extend it in the future. This is a very similar model to how `Pod.spec.containers.env` is handled, so we can probably borrow some decisions from there. We probably also want to allow fieldRefs as well as label references here too.

PR #247, opened 2018-02-12
WIP: Cassandra: Configure GossipingPropertyFileSnitch on pilot startup

kragniz:
This switches cassandra to use `GossipingPropertyFileSnitch` as the snitch. This takes input from a configuration file about the failure zones the cassandra node lives in and gossips it to other nodes in the cluster. Currently the inputs are the node annotations `failure-domain.beta.kubernetes.io/region` and `failure-domain.beta.kubernetes.io/zone` as the datacenter and rack respectively. 

Note: this is temporarily based off #245 while we work out how to handle extra configuration settings to `cassandra.yaml`.

```release-note
Teach cassandra about the kubernetes failure zones.
```


pkg/controllers/cassandra/role/control.go
```diff
@@ -71,6 +71,11 @@ func RoleForCluster(cluster *v1alpha1.CassandraCluster) *rbacv1.Role {
 				Verbs:     []string{"create", "update", "patch", "get", "list", "watch"},
 				Resources: []string{"configmaps"},
 			},
+			{
+				APIGroups: []string{""},
+				Verbs:     []string{"get", "list", "watch"},
+				Resources: []string{"endpoints"},
```
munnerz: `s/endpoints/nodes`?

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -39,9 +45,65 @@ func NewPilot(opts *PilotOptions) (*Pilot, error) {
...
+		return nil, err
+	}
+
+	newContents := strings.Replace(string(read),
+		"org.apache.cassandra.locator.SimpleSeedProvider",
+		"io.k8s.cassandra.KubernetesSeedProvider", -1)
+
+	newContents = strings.Replace(string(newContents),
+		"SimpleSnitch",
+		"GossipingPropertyFileSnitch", -1)
```
munnerz: lgtm for now until we invent better config handling constructs

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -39,9 +45,65 @@ func NewPilot(opts *PilotOptions) (*Pilot, error) {
...
+		return nil, err
+	}
+
+	snitchSettings, err := getSnitchSettings(opts.kubeClientset)
+	if err != nil {
+		return nil, err
+	}
+
+	glog.V(4).Info("generated snitch:")
+	glog.V(4).Info(snitchSettings)
```
munnerz: collapse into a single call to print else other log entries could be interlaced with this: `glog.V(4).Infof("generated snitch config: %v", snitchSettings)`

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -39,9 +45,65 @@ func NewPilot(opts *PilotOptions) (*Pilot, error) {
...
+	}
+
 	return p, nil
 }
 
+func getSnitchSettings(clientset kubernetes.Interface) (string, error) {
+	nodeClientset := clientset.CoreV1().Nodes()
+
+	nodeName := os.Getenv("NODE_NAME")
+	node, err := nodeClientset.Get(nodeName, metav1.GetOptions{})
```
munnerz: So making this call is something that's blocked a similar PR on the ES pilot. I'm not sure if we want to allow Pilots to `get nodes`. I'll need to dig out the literature, but exposing node information to pods poses new security considerations as the pod is able to infer more information about where it is running.

I'd be happy to run with this for now until we hear more, but nodes are non-namespaced resources etc. so this is a more considerable change than first meets the eye. That said, I think we'd need some gRPC/synchronous (and highly available) distribution point for this configuration otherwise, which doesn't seem like the *first* thing we should rush to try.

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -39,9 +45,65 @@ func NewPilot(opts *PilotOptions) (*Pilot, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	datacenter := "navigator-datacenter-1"
+	rack := "navigator-rack-1"
```
munnerz: What if these are left blank? Is blank better than a default?

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -39,9 +45,65 @@ func NewPilot(opts *PilotOptions) (*Pilot, error) {
...
+
+	if val, ok := node.Annotations["failure-domain.beta.kubernetes.io/region"]; ok {
+		datacenter = val
+	}
+
+	if val, ok := node.Annotations["failure-domain.beta.kubernetes.io/zone"]; ok {
+		rack = val
+	}
```
munnerz: I don't think we can enforce that zone and region are the hardcoded keys we support - there may be some environments where e.g. they want to distribute over many physical RAID arrays, in which case being able to use arbitrary node labels here becomes valuable.

There is a similar feature in Elasticsearch, 'Shard Allocation Awareness' - the only difference with that is it supports more than just DC/Region/Rack (you can add arbitrary 'keys' that can then be reasoned over). I wrote up a small proposal as to the API surface of this here: https://github.com/jetstack/navigator/issues/44#issuecomment-357212909

wdyt? I like the "out of the box it's replicated" behaviour, but should we expose it through a more advanced syntax that has defaults?

pkg/pilot/cassandra/v3/pilot.go
kragniz: It'd be interesting to see those security worries. This sort of information about node placement is probably going to be necessary for most dbs, so maybe a minimal set of it could be located in the navigator api?

munnerz: Yep this is possible, but I'd rather keep it as-is for now. Copying that information across to Navigator could become more nuanced than we initially think - e.g. what if stale data exists on the Pilot resource, or if the data doesn't already exist?

I too am not certain on those security worries - let's not address it until we need to/we've got feedback from users 😄

pkg/pilot/cassandra/v3/pilot.go
munnerz: Realistically we won't be able to make it *that* general purpose (imo). Notably due to the limitations we're already seeing in different DB types.

That said, the use of the two should be *similar*. If someone has been operating Navigator ES databases for some time, they should be able to pick up the Cassandra resource and know how to apply those rules there. (i.e. field names might be different, but field values/types should be similar/the same)

pkg/pilot/cassandra/v3/pilot.go
kragniz: I think they then get assigned to the cassandra defaults (`rack1` in `datacenter1`). I picked a different default so we could see if GossipingPropertyFileSnitch is enabled (if we see nodes assigned to `rack1`, we know something has gone wrong)

pkg/pilot/cassandra/v3/pilot.go
kragniz: Another option to avoid giving access to `get nodes` is to send this information via labels and the downward api: We could watch for pilot pods and label them with the appropriate failure zones from the node once they're scheduled. This could then be exposed in the pod using `fieldPath: metadata.labels` from the downward api.

We'd have to wait for those labels to get propagated into the pod before starting cassandra, so it might cause too much latency when starting a cluster.

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -94,6 +94,14 @@ func StatefulSetForCluster(
 									jolokiaPort,
 									jolokiaContext,
 								),
+								fmt.Sprintf(
+									"--datacenter-node-label=%s",
+									np.FailureZones.Datacenter.ValueFrom.NodeLabel,
```
kragniz: We should actually check this exists, and pick between `Value`, or one of the potential `ValueFrom`s

kragniz:
Superceded by #255

PR #249, opened 2018-02-13
Add apiserver.persistence config stanza to helm chart

### LINKED ISSUE #135: Support adding persistent volume to navigator-apiserver etcd container
/kind bug

* I had a cluster with Navigator databases running.
* I scaled the cluster down to 0 nodes.
* When I scale the cluster back up, the Navigator resources are all missing.
* And the StatefulSets and services that Navigator controller created are all gone too.

munnerz:

Allows persisting apiserver state with a pvc.

**Which issue this PR fixes**: fixes #135 

**Release note**:
```release-note
Add option to use PVC for persisting apiserver etcd data
```

### DIFF
--- a/contrib/charts/navigator/templates/apiserver-pvc.yaml
+++ b/contrib/charts/navigator/templates/apiserver-pvc.yaml
@@ -1,0 +1,18 @@
+{{- if .Values.apiserver.persistence.enabled }}
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{ template "fullname" . }}-apiserver
+  annotations:
+  {{- if .Values.apiserver.persistence.storageClass }}
+    volume.beta.kubernetes.io/storage-class: {{ .Values.apiserver.persistence.storageClass | quote }}
+  {{- else }}
+    volume.alpha.kubernetes.io/storage-class: default
+  {{- end }}
+spec:
+  accessModes:
+    - {{ .Values.apiserver.persistence.accessMode | quote }}
+  resources:
+    requests:
+      storage: {{ .Values.apiserver.persistence.size | quote }}
+{{- end }}
--- a/contrib/charts/navigator/values.yaml
+++ b/contrib/charts/navigator/values.yaml
@@ -25,6 +25,15 @@ apiserver:
     tag: latest
     pullPolicy: Always
   logLevel: 2
+
+  persistence:
+    enabled: false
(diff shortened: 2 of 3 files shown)


munnerz:
In future we should expand the helm chart to support specifying a custom etcd url to connect to, in order to support HA apiserver configurations (which should be the recommended deployment method)

contrib/charts/navigator/templates/apiserver-pvc.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.apiserver.persistence.enabled }}
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{ template "fullname" . }}-apiserver
+  annotations:
+  {{- if .Values.apiserver.persistence.storageClass }}
+    volume.beta.kubernetes.io/storage-class: {{ .Values.apiserver.persistence.storageClass | quote }}
+  {{- else }}
+    volume.alpha.kubernetes.io/storage-class: default
```
wallrj: Why `alpha` and `beta` in these two branches?

contrib/charts/navigator/values.yaml
```diff
@@ -25,6 +25,15 @@ apiserver:
     tag: latest
     pullPolicy: Always
   logLevel: 2
+
+  persistence:
+    enabled: false
+    size: 10Gi
+    accessMode: ReadWriteOnce
```
wallrj: Why is accessMode configurable?  `ReadWriteOnce` seems like the only sensible option.

contrib/charts/navigator/templates/apiserver-pvc.yaml
wallrj: Ah, I see that the "default" storage class name is an alpha feature.
Need to fix that in the cassandra chart then.

munnerz:
Re: whether we should not include any provisions for local etcd at all:

Okay sure, but then we need to remove *all* support for running etcd for Navigator and instead just expose a `--etcd-servers`, `client-cert` etc configuration option in the chart. (e2e tests will then also need appropriately updating to run etcd).

Given the early nature of Navigator, and the fact a user can run `etcdctl backup` in order to snapshot etcd (in case they want to switch to a clustered version), I figured this was a good stop gap. Currently the only way someone can persist data is to fork our Helm chart.

PR #242, opened 2018-02-06
Add ginkgo e2e testing framework

munnerz:
**What this PR does / why we need it**:

Adds a ginkgo based e2e testing framework, and updates the Elasticsearch e2e tests to be written with it.

fixes #32, fixes #156

**Special notes for your reviewer**:

Stacked on #241, #240 and #239 

**Release note**:
```release-note
NONE
```

PR #254, opened 2018-02-20
Mark the first pod in a cassandra nodepool as a seed

kragniz:
This is an alternative to #252 that marks the first pod in a nodepool (rather than the first pod in each failure zone) as a seed, the intention being we'd have a nodepool per region to add fault tolerance.


pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -68,6 +79,61 @@ func (e *defaultCassandraClusterNodepoolControl) removeUnusedStatefulSets(
...
+	var nodepoolPods []*v1.Pod
+	allPods, err := e.pods.Pods(cluster.Namespace).List(labels.Everything())
+	if err != nil {
+		return err
+	}
+	for _, pod := range allPods {
+		if metav1.IsControlledBy(pod, set) {
+			nodepoolPods = append(nodepoolPods, pod)
+		}
+	}
```
wallrj: :grey_question: This should work, but we also add nodepool labels to the pods, so it might also be possible to supply a more specific label selector to the lister, above.

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -68,6 +79,61 @@ func (e *defaultCassandraClusterNodepoolControl) removeUnusedStatefulSets(
...
+		elements := strings.Split(p.Name, "-")
+		n, err := strconv.Atoi(elements[len(elements)-1])
+		if err != nil {
+			return err
+		}
+		numberedPods[n] = p
+		ns = append(ns, n)
+	}
+
+	pod := numberedPods[min(ns)]
```
wallrj: :grey_question: There are some useful utilities (`getOrdinal` and `[]*v1.Pod` sorted by ordinal) in https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/statefulset/stateful_set_utils.go but they're private :cry: 

I also found this issue (which may have been addressed) for exposing the pod ordinal as a  pod annotation.
 * https://github.com/kubernetes/kubernetes/pull/30860

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -68,6 +79,61 @@ func (e *defaultCassandraClusterNodepoolControl) removeUnusedStatefulSets(
...
+	pod.Labels["seed"] = "true"
+	e.kubeClient.CoreV1().Pods(pod.Namespace).Update(pod)
```
wallrj: Elsewhere, we take special care not to mutate the objects returned by informers / listers. 
Doing so will change the object in the shared cache.
Instead look at making a `DeepCopy`.

wallrj: Also need to handle the error returned by the update.

wallrj: And probably should only perform the update if the pod does not already have the label.

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -87,6 +153,9 @@ func (e *defaultCassandraClusterNodepoolControl) createOrUpdateStatefulSet(
 	if err != nil {
 		return err
 	}
+
+	e.labelSeedNodes(cluster, existingSet)
```
wallrj: Handle the error here.

pkg/controllers/cassandra/service/seedprovider/resource.go
```diff
@@ -27,6 +27,10 @@ func updateServiceForCluster(
 	service.SetName(util.SeedProviderServiceName(cluster))
 	service.Spec.Type = apiv1.ServiceTypeClusterIP
 	service.Spec.ClusterIP = "None"
+
+	// Only mark nodes explicitly labeled as seeds as seed nodes
+	service.Spec.Selector["seed"] = "true"
+
```
wallrj: :grey_question: Sorry about my over complicated code here :-( I was trying to share as much between the two services as possible, hence the `SetStandardServiceAttributes` stuff. 
But now I'm thinking that the CQL service probably isn't necessary and all this could be simplified in a future branch.

pkg/controllers/cassandra/nodepool/nodepool.go
munnerz: There's some parts of our codebase already (it may be as part of e2e test utils, or validation functions) for calculating the ordinal of a pod in a statefulset.

munnerz: ... if it is in e2e test utils, that'll be as part of the ginkgo PR which hasn't merged yet come to think #242

munnerz: `min` currently panics if the slice passed to it is len 0 and we don't check the length of ns here.

`min` itself should probably be changed (I'd be happy to have it just return 0, and add a comment to explain this behaviour).

munnerz:
@wallrj 

> And I guess we need to be clear that this has some (serious) limitations...that if pod-0 is restarted after the cluster has been established, it'll be looking at itsself as seed provider and go nuts....which is probably what would have happened pre-kubernernetes seed provider.

:thinking: that also implies we don't support single node cassandra clusters? Is the problem here that the seed node won't be 'Ready' so won't be available via the seed node service? If so, we should probably 'tolerate unready endpoints' on the seed service (we should probably do this regardless tbh, as I guess the cassandra clients will attempt to connect to each seed node in turn if one fails)

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -68,6 +79,61 @@ func (e *defaultCassandraClusterNodepoolControl) removeUnusedStatefulSets(
...
+	ns := []int{}
+	for _, p := range nodepoolPods {
+		elements := strings.Split(p.Name, "-")
+		n, err := strconv.Atoi(elements[len(elements)-1])
+		if err != nil {
+			return err
+		}
+		numberedPods[n] = p
+		ns = append(ns, n)
+	}
```
munnerz: By selecting the lowest number pod, we'll be changing which pod is used as a seed every time the 0th pod fails.

Instead, we should *always* use `fmt.Sprintf("%s-0", p.Name)`.

pkg/controllers/cassandra/nodepool/nodepool.go
munnerz: (also just noticed you do actually check the length of `nodepoolPods`, but as I said in my other comment I think the min function can go altogether as we always want to use -0)

pkg/controllers/cassandra/nodepool/nodepool.go
kragniz: :+1: doing it this way makes it much simpler, I've switched the latest version to this

PR #255, opened 2018-02-21
Add datacenter and rack info to cassandra nodepool spec

kragniz:
This adds three fields to cassandra's nodepool spec:
- `nodeSelector`
- `rack`
- `datacenter`

`nodeSelector` is passed to the statefulset spec, allowing users to specify where the nodepool is scheduled.

`rack` and `datacenter` are both passed to cassandra's GossipingPropertyFileSnitch, which distributes this information around the rest of the cluster.

Fixes #247 

```release-note
Add datacenter and rack info to cassandra nodepool spec
```

### DIFF
--- a/pkg/apis/navigator/v1alpha1/types.go
+++ b/pkg/apis/navigator/v1alpha1/types.go
@@ -31,7 +31,7 @@ type CassandraClusterSpec struct {
 	NavigatorClusterConfig `json:",inline"`
 
 	NodePools []CassandraClusterNodePool `json:"nodePools"`
-	Image     ImageSpec                  `json:"image"`
+	Image     *ImageSpec                 `json:"image"`
 	CqlPort   int32                      `json:"cqlPort"`
 }
 
--- a/pkg/apis/navigator/v1alpha1/types.go
+++ b/pkg/apis/navigator/v1alpha1/types.go
@@ -49,7 +49,14 @@ type CassandraClusterNodePool struct {
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector"`
 
-	Rack       string `json:"rack"`
+	// Rack specifies the cassandra rack with which to label nodes in this
+	// nodepool. If this is not set, a default will be selected.
+	// +optional
+	Rack string `json:"rack"`
+
+	// Datacenter specifies the cassandra datacenter with which to label nodes
+	// in this nodepool. If this is not set, a default will be selected.
+	// +optional
 	Datacenter string `json:"datacenter"`
 }
 
--- a/pkg/apis/navigator/v1alpha1/types.go
+++ b/pkg/apis/navigator/v1alpha1/types.go
@@ -43,6 +43,11 @@ type CassandraClusterNodePool struct {
 	// Persistence specifies the configuration for persistent data for this
 	// node.
 	Persistence PersistenceConfig `json:"persistence,omitempty"`
+
+	// NodeSelector should be specified to force nodes in this pool to run on
+	// nodes matching the given selector.
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector"`
 }
 
 type CassandraClusterStatus struct {
--- a/pkg/apis/navigator/v1alpha1/types.go
+++ b/pkg/apis/navigator/v1alpha1/types.go
@@ -31,7 +31,7 @@ type CassandraClusterSpec struct {
 	NavigatorClusterConfig `json:",inline"`
 
 	NodePools []CassandraClusterNodePool `json:"nodePools"`
-	Image     *ImageSpec                 `json:"image"`
+	Image     ImageSpec                  `json:"image"`
 	CqlPort   int32                      `json:"cqlPort"`
 }
 
--- a/pkg/apis/navigator/v1alpha1/types.go
+++ b/pkg/apis/navigator/v1alpha1/types.go
@@ -48,6 +48,9 @@ type CassandraClusterNodePool struct {
 	// nodes matching the given selector.
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector"`
+
+	Rack       string `json:"rack"`
+	Datacenter string `json:"datacenter"`
 }
 
 type CassandraClusterStatus struct {
--- a/pkg/controllers/cassandra/nodepool/resource.go
+++ b/pkg/controllers/cassandra/nodepool/resource.go
@@ -21,6 +21,8 @@ const (
 	cassDataVolumeName      = "cassandra-data"
 	cassDataVolumeMountPath = "/var/lib/cassandra"
 
+	cassSnitch = "GossipingPropertyFileSnitch"
+
 	// See https://jolokia.org/reference/html/agents.html#jvm-agent
 	jolokiaHost    = "127.0.0.1"
 	jolokiaPort    = 8778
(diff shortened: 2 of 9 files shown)


pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -43,6 +43,14 @@ type CassandraClusterNodePool struct {
...
 	// node.
 	Persistence PersistenceConfig `json:"persistence,omitempty"`
+
+	// NodeSelector should be specified to force nodes in this pool to run on
+	// nodes matching the given selector.
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector"`
+
+	Rack       string `json:"rack"`
+	Datacenter string `json:"datacenter"`
```
munnerz: Needs a comment to explain that if these aren't set, a default rack and datacenter will be specified.
Needs +optional on the fields.

munnerz: Also, are we going to auto-populate the `zone` or `datacenter` fields if the node selector includes a selector on zone?

If so, we need to add that functionality and a comment here explaining as such.

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -35,6 +35,17 @@ func StatefulSetForCluster(
...
+
+	datacenter := "navigator-default-datacenter"
+	if np.Datacenter != "" {
+		datacenter = np.Datacenter
+	}
+
+	rack := "navigator-default-rack"
+	if np.Rack != "" {
+		rack = np.Rack
+	}
```
munnerz: Would it make sense to default 'rack' to the node pool name? The only problem with that is it requires explicit action from the user to *disable* the rack awareness (they'd need to explicitly set 'rack' to some static value on all node pools)

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -188,6 +200,10 @@ func StatefulSetForCluster(
 										cluster.Namespace,
 									),
 								},
+								{
+									Name:  "CASSANDRA_ENDPOINT_SNITCH",
+									Value: "GossipingPropertyFileSnitch",
```
munnerz: Perhaps make this a const so it's easy to see which snitch we use?

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -35,6 +35,17 @@ func StatefulSetForCluster(
 	statefulSetName := util.NodePoolResourceName(cluster, np)
 	seedProviderServiceName := util.SeedProviderServiceName(cluster)
 	nodePoolLabels := util.NodePoolLabels(cluster, np.Name)
+
+	datacenter := "navigator-default-datacenter"
```
munnerz: default datacenter & rack should probably also be a const in this file

pkg/controllers/cassandra/nodepool/resource.go
kragniz: :+1:

pkg/controllers/cassandra/nodepool/resource.go
kragniz: This sounds reasonable. I'd imagine if a user didn't care about rack awareness they would only specify a single nodepool resource.

pkg/apis/navigator/v1alpha1/types.go
kragniz: I think it'd be better to auto-populate the zone based on the nodepool name (as you mentioned below), and the region as some static default.

pkg/controllers/cassandra/nodepool/resource.go
munnerz: I'm not 100% sold though, as it may be surprising to a user that has never specified a `rack`, `datacenter` or `nodeSelector` including a zone attribute to find that their data is being spread in a rack-aware fashion.

Spreading across node pools by default isn't really that desirable if the properties of that node pool are relatively similar (e.g. spread across all zones in the cluster due to lack of nodeSelector).

cc @wallrj

wallrj: I'm not really sure.
 * I think using nodepool name as rack name is reasonable until we understand it better.
 * If you don't specify a `nodeSelector` the nodepool pods and their PVs could end up in any zone. And once those PVs are created it probably means that the pod is stuck in that zone (on AWS and GCP at least). 
So the nodeselector should probably be immutable too.

wallrj: In fact my preference would be to remove the defaults and make these required fields....but I understand if you decide to go with the defaults for now.

kragniz:
In the latest patches, I've added validation of the new fields (as well as some additional validation) and set the default rack value to the name of the nodepool

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -35,6 +39,17 @@ func StatefulSetForCluster(
 	statefulSetName := util.NodePoolResourceName(cluster, np)
 	seedProviderServiceName := util.SeedProviderServiceName(cluster)
 	nodePoolLabels := util.NodePoolLabels(cluster, np.Name)
+
+	datacenter := cassDefaultDatacenter
+	if np.Datacenter != "" {
+		datacenter = np.Datacenter
+	}
```
wallrj: :grey_question: Not suggesting now, but might we also create a (mutating) AdmissionController to do this? Ala https://kubernetes.io/docs/admin/admission-controllers/#defaultstorageclass

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -35,6 +39,17 @@ func StatefulSetForCluster(
...
+
+	rack := np.Name
+	if np.Rack != "" {
+		rack = np.Rack
+	}
+
```
wallrj: :grey_question: Does Cassandra place any limit on the length of or characters in  this field? `np.Name` can be any DNS label (I think).

pkg/registry/navigator/cassandracluster/strategy.go
```diff
@@ -89,8 +89,10 @@ func (cassandraClusterStrategy) Canonicalize(obj runtime.Object) {
 }
 
 func (cassandraClusterStrategy) ValidateUpdate(ctx genericapirequest.Context, obj, old runtime.Object) field.ErrorList {
-	cass := obj.(*navigator.CassandraCluster)
-	return validation.ValidateCassandraCluster(cass)
+	newCassandraCluster := obj.(*navigator.CassandraCluster)
+	oldCassandraCluster := old.(*navigator.CassandraCluster)
+
+	return validation.ValidateCassandraClusterUpdate(oldCassandraCluster, newCassandraCluster)
```
wallrj: :grey_question: I'm still unclear what happens when I do `kubectl replace --file my/cassandra/manifest.yaml` Will that bypass this update check? Or only when I do `kubectl replace --force` ?

pkg/controllers/cassandra/nodepool/resource.go
kragniz: From what I can tell from cassandra's source, the maximum is the maximum size of a java string (big)

pkg/registry/navigator/cassandracluster/strategy.go
kragniz: It doesn't seem to bypass it:
```
 g/s/g/j/navigator {cassandra-multiple-nodepools} $ k replace -f cassandra.yaml                             
The CassandraCluster "mydb" is invalid: spec.nodePools[0].datacenter: Forbidden: cannot modify datacenter   
```

with --force, it gets accepted though:
```
 g/s/g/j/navigator {cassandra-multiple-nodepools} $ k replace -f cassandra.yaml --force                                                                                                                                                                                                                                  407ms 
cassandracluster "mydb" deleted
cassandracluster "mydb" replaced
```

Issue #253, opened 2018-02-16
Refactor Cassandra cluster controller into Actions


wallrj:
Implement CreateCluster and ScaleUp state change actions for Cassandra.

Based on https://github.com/jetstack/navigator/pull/241

/kind feature

PR #256, opened 2018-02-22
Add CreateNodePool and ScaleOut actions for Cassandra

wallrj:
These become the only changes supported by the Cassandra controller until ScaleIn and  CassandraUpgrade actions are implemented in followup branches.

Fixes: #253 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/internal/test/util/generate/generate.go
+++ b/internal/test/util/generate/generate.go
@@ -118,3 +118,28 @@ func StatefulSet(c StatefulSetConfig) *apps.StatefulSet {
 		},
 	}
 }
+
+type CassandraClusterConfig struct {
+	Name, Namespace string
+}
+
+func CassandraCluster(c CassandraClusterConfig) *v1alpha1.CassandraCluster {
+	return &v1alpha1.CassandraCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      c.Name,
+			Namespace: c.Namespace,
+		},
+	}
+}
+
+type CassandraClusterNodePoolConfig struct {
+	Name     string
+	Replicas int32
+}
+
+func CassandraClusterNodePool(c CassandraClusterNodePoolConfig) *v1alpha1.CassandraClusterNodePool {
+	return &v1alpha1.CassandraClusterNodePool{
+		Name:     c.Name,
+		Replicas: c.Replicas,
+	}
+}
--- a/pkg/controllers/cassandra/actions/create_nodepool.go
+++ b/pkg/controllers/cassandra/actions/create_nodepool.go
@@ -1,0 +1,39 @@
+package actions
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
+	"github.com/jetstack/navigator/pkg/controllers"
+	"github.com/jetstack/navigator/pkg/controllers/cassandra/nodepool"
+)
+
+type CreateNodePool struct {
+	Cluster  *v1alpha1.CassandraCluster
+	NodePool *v1alpha1.CassandraClusterNodePool
+}
+
+var _ controllers.Action = &CreateNodePool{}
+
+func (a *CreateNodePool) Name() string {
+	return "CreateNodePool"
+}
+
+func (a *CreateNodePool) Execute(s *controllers.State) error {
+	ss := nodepool.StatefulSetForCluster(a.Cluster, a.NodePool)
+	_, err := s.Clientset.AppsV1beta1().StatefulSets(ss.Namespace).Create(ss)
+	if k8sErrors.IsAlreadyExists(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	s.Recorder.Eventf(
+		a.Cluster,
+		corev1.EventTypeNormal,
+		a.Name(),
+		"CreateNodePool: Name=%q", a.NodePool.Name,
+	)
+	return nil
+}
--- a/pkg/controllers/cassandra/actions/scaleout.go
+++ b/pkg/controllers/cassandra/actions/scaleout.go
@@ -1,0 +1,59 @@
+package actions
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/golang/glog"
+
+	"github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
+	"github.com/jetstack/navigator/pkg/controllers"
+	"github.com/jetstack/navigator/pkg/controllers/cassandra/nodepool"
+	"github.com/jetstack/navigator/pkg/util/ptr"
+)
+
+type ScaleOut struct {
+	Cluster  *v1alpha1.CassandraCluster
+	NodePool *v1alpha1.CassandraClusterNodePool
+}
+
+var _ controllers.Action = &ScaleOut{}
+
+func (a *ScaleOut) Name() string {
+	return "ScaleOut"
+}
+
+func (a *ScaleOut) Execute(s *controllers.State) error {
+	baseSet := nodepool.StatefulSetForCluster(a.Cluster, a.NodePool)
+	existingSet, err := s.StatefulSetLister.
+		StatefulSets(baseSet.Namespace).Get(baseSet.Name)
+	if err != nil {
+		return err
+	}
+	newSet := existingSet.DeepCopy()
+	if *existingSet.Spec.Replicas == a.NodePool.Replicas {
+		return nil
+	}
+	if *existingSet.Spec.Replicas > a.NodePool.Replicas {
+		glog.Errorf(
+			"ScaleOut error:"+
+				"The StatefulSet.Spec.Replicas value (%d) "+
+				"is greater than the desired value (%d)",
+			*existingSet.Spec.Replicas, a.NodePool.Replicas,
+		)
+		return nil
+	}
+	newSet.Spec.Replicas = ptr.Int32(*newSet.Spec.Replicas + 1)
+	_, err = s.Clientset.AppsV1beta1().
+		StatefulSets(newSet.Namespace).Update(newSet)
+	if err != nil {
+		return err
+	}
+	s.Recorder.Eventf(
+		a.Cluster,
+		corev1.EventTypeNormal,
+		a.Name(),
+		"ScaleOut: NodePool=%q, ReplicaCount=%d, TargetReplicaCount=%d",
+		a.NodePool.Name, *newSet.Spec.Replicas, a.NodePool.Replicas,
+	)
+	return nil
+}
--- a/pkg/controllers/cassandra/cluster_control.go
+++ b/pkg/controllers/cassandra/cluster_control.go
@@ -6,6 +6,8 @@ import (
 	"k8s.io/client-go/tools/record"
 
 	v1alpha1 "github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
+	"github.com/jetstack/navigator/pkg/controllers"
+	"github.com/jetstack/navigator/pkg/controllers/cassandra/actions"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra/nodepool"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra/pilot"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra/role"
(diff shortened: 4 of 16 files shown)


wallrj:
rebased

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -44,7 +44,7 @@ type CassandraClusterSpec struct {
 // CassandraClusterNodePool describes a node pool within a CassandraCluster.
 type CassandraClusterNodePool struct {
 	Name     string `json:"name"`
-	Replicas int64  `json:"replicas"`
+	Replicas int32  `json:"replicas"`
```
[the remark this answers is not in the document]
wallrj: I changed it to match the type of `StatefulSet.Status.ReadyReplicas`.
 * https://github.com/kubernetes/kubernetes/blob/v1.9.3/pkg/apis/apps/types.go#L171

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -34,77 +36,65 @@ func NewControl(
...
 	cluster *v1alpha1.CassandraCluster,
-) error {
-	expectedStatefulSetNames := map[string]bool{}
-	for _, pool := range cluster.Spec.NodePools {
-		name := util.NodePoolResourceName(cluster, &pool)
-		expectedStatefulSetNames[name] = true
-	}
-	client := e.kubeClient.AppsV1beta1().StatefulSets(cluster.Namespace)
+) (results map[string]*v1beta1.StatefulSet, err error) {
+	results = map[string]*v1beta1.StatefulSet{}
```
munnerz: I'm confused what this package does now that we've switched to Actions for managing node pools? When I implemented this in the ES controller, the `nodepool` package was eventually refactored away entirely.

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -49,7 +48,7 @@ func StatefulSetForCluster(
 			OwnerReferences: []metav1.OwnerReference{util.NewControllerRef(cluster)},
 		},
 		Spec: apps.StatefulSetSpec{
-			Replicas:    util.Int32Ptr(int32(np.Replicas)),
+			Replicas:    util.Int32Ptr(int32(0)),
```
munnerz: 0?

pkg/apis/navigator/v1alpha1/gen.go
```diff
@@ -0,0 +1,70 @@
...
+			nodepools[i] = v.Interface().(CassandraClusterNodePool)
+		}
+	}
+	for _, np := range nodepools {
+		o.NodePools[np.Name] = CassandraClusterNodePoolStatus{
+			ReadyReplicas: np.Replicas,
+		}
+	}
+	return reflect.ValueOf(o)
+}
```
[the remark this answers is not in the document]
wallrj: That's the interface used by quick/test to generate test values.
I don't think you can add methods to a struct from a separate package.
Since was a bit of an experiment I'll move it to a separate branch for now.
Also want to try using go-fuzz and gopter at some point.

pkg/controllers/cassandra/actions/scaleout.go
```diff
@@ -0,0 +1,42 @@
...
+		ss.Spec.Replicas = &a.NodePool.Replicas
+		_, err = s.Clientset.AppsV1beta1().StatefulSets(ss.Namespace).Update(ss)
+		return err
+	}
+	if *ss.Spec.Replicas > a.NodePool.Replicas {
+		return fmt.Errorf(
+			"the NodePool.Replicas value (%d) "+
+				"is less than the existing StatefulSet.Replicas value (%d)",
+			a.NodePool.Replicas, *ss.Spec.Replicas,
+		)
```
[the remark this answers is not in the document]
wallrj: It is already in the next action method.
Duplicated here in case the state of the world has changed between the action being chosen and being executed.

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -162,13 +170,51 @@ func (e *defaultCassandraClusterControl) Sync(c *v1alpha1.CassandraCluster) erro
...
 	e.recorder.Event(
 		c,
 		apiv1.EventTypeNormal,
 		SuccessSync,
 		MessageSuccessSync,
 	)
+	return c.Status, nil
+}
+
+func NextAction(c *v1alpha1.CassandraCluster) controllers.Action {
```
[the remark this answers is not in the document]
wallrj: I'd like to see how far I can get relying only on the `CassandraCluster.Status`.

pkg/controllers/cassandra/nodepool/nodepool.go
wallrj: I re-purposed it to be responsible only for updating the `CassandraCluster.Status.NodePools`.

pkg/controllers/cassandra/nodepool/resource.go
wallrj: So that NodePools are always created with size 0 and then scaled out according to the logic in `ScaleOut.Execute`.

pkg/controllers/cassandra/nodepool/nodepool.go
munnerz: Given the above model of only deciding which action to perform next based on the Status field, can we move this into a `UpdateClusterStatus` function in the `cassandra` package? I like the idea of keeping this sort of info centralised, and it saves having the extra package (which given other similar packages e.g. `service` create resources, may be confusing in future).

*all* changes to the cluster status can then be computed and set in this function.

pkg/controllers/cassandra/nodepool/resource.go
munnerz: Cool 👍

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -34,77 +36,65 @@ func NewControl(
...
+		nodePoolNames.Insert(np.Name)
+		ssName := util.NodePoolResourceName(cluster, &np)
+		ss, setFound := ssList[ssName]
+		nps, npsFound := cluster.Status.NodePools[np.Name]
+		if setFound {
+			if !npsFound {
+				cluster.Status.NodePools[np.Name] = nps
+			}
+			if nps.ReadyReplicas != ss.Status.ReadyReplicas {
+				nps.ReadyReplicas = ss.Status.ReadyReplicas
```
munnerz: It took a little while to spot that this is where we actually set the fields in the node pool status - perhaps create a function `statusForNodePool` to handle this? (which returns a NodePoolStatus structure)

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -34,77 +36,65 @@ func NewControl(
...
-		return err
+// Add a NodePoolStatus for each NodePool, only if a corresponding StatefulSet is found.
+// Update the NodePoolStatus for each NodePool, using values from the corresponding StatefulSet.
+// Remove the NodePoolStatus for NodePools that do not have a StatefulSet
+// (the statefulset has been deleted unexpectedly)
+// Remove the NodePoolStatus if there is no corresponding NodePool.
+// (the statefulset has been removed by a DeleteNodePool action - not yet implemented)
+func (e *defaultCassandraClusterNodepoolControl) Sync(cluster *v1alpha1.CassandraCluster) error {
+	if cluster.Status.NodePools == nil {
+		cluster.Status.NodePools = map[string]v1alpha1.CassandraClusterNodePoolStatus{}
```
munnerz: Not 100% sure how maps are ordered when serialised - perhaps we shouldn't be using a map here at all in favour of a slice that has a `name` field?

pkg/controllers/cassandra/testing/gen.go
```diff
@@ -0,0 +1,46 @@
...
+		np := v1alpha1.CassandraClusterNodePool{
+			Name: fmt.Sprintf("np%d", i),
+		}
+		FuzzCassandraNodePool(&np, rand, size)
+		nps := v1alpha1.CassandraClusterNodePoolStatus{
+			ReadyReplicas: np.Replicas,
+		}
+		// 20% chance of ScaleOut
+		if rand.Intn(4) == 0 {
+			np.Replicas++
```
munnerz: Will increase it by 1 ever result in it being particularly large? Should we not just set this to some random value output by the rng?

munnerz: (maybe you are following some best practice I am not aware of though!)

wallrj:
Noticed a weird E2E test error:
```
W0324 14:27:19.643] + kubectl --namespace=test-cassandra-1521900884-833 exec cass-test-ringnodes-0 -- bash -c 'kill -SIGSTOP -- $(ps -u cassandra -o pid=) && ps faux'
I0324 14:27:20.340] rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:262: starting container process caused "process_linux.go:86: adding pid 17816 to cgroups caused \"failed to write 17816 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/kubepods/poddc1388be-2f6e-11e8-8210-5254001464f2/f148ee0881cb5644368505fc324414543c01c6046f6e9850c52cc4f48ce72801/cgroup.procs: invalid argument\""
```
 * https://jetstack-build-infra.appspot.com/build/jetstack-logs/pr-logs/pull/jetstack_navigator/256/navigator-e2e-v1-7/1071/

Maybe caused by https://github.com/opencontainers/runc/issues/1326 or https://github.com/moby/moby/issues/31230

pkg/controllers/cassandra/nodepool/nodepool.go
wallrj: I'm going to leave it like this for now.
My memory of previous conversations about the ``Controller.Sync`` pattern is that the ``Sync`` method is expected to mutate the ``CassandraCluster.Status`` and having run all the `control` s, the `Controller` performs an ``UpdateStatus`` with the final status.

Having it in a separate package makes it easier to unit test the NodePoolStatus behaviour in isolation.

pkg/controllers/cassandra/actions/create_nodepool.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (a *CreateNodePool) Name() string {
+	return "CreateNodePool"
+}
+
+func (a *CreateNodePool) Execute(s *controllers.State) error {
+	ss := nodepool.StatefulSetForCluster(a.Cluster, a.NodePool)
+	_, err := s.Clientset.AppsV1beta1().StatefulSets(ss.Namespace).Create(ss)
+	if k8sErrors.IsAlreadyExists(err) {
+		return nil
```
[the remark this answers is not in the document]
wallrj: There are a few good reasons for them to be idempotent:
 1. the `CassandraCluster.Status.NodePools` map (upon which this action is calculated) may not have been updated.
 2. future actions (e.g. ScaleIn) may perform a more complex sequence of steps which need to be "ratcheted" through, by repeated calls to `ScaleIn.Execute`, e.g.
   1. Decommission node
   2. Decrease the statefulset replica count.
  In which case each of those intermediate steps must be idempotent and in which case the final step may as well be idempotent.
 3. The lister used to calculate the action may not have the most up-to-date  resources, so it appears there isn't a StatefulSet, until you attempt to create it, which fails with AlreadyExists (although I might be making that up, because the current implementation doesn't check the lister first)

On the whole, it makes sense to me that these actions should all be idempotent, but happy to discuss further.

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -34,77 +36,65 @@ func NewControl(
...
+			if !npsFound {
+				cluster.Status.NodePools[np.Name] = nps
+			}
+			if nps.ReadyReplicas != ss.Status.ReadyReplicas {
+				nps.ReadyReplicas = ss.Status.ReadyReplicas
+				cluster.Status.NodePools[np.Name] = nps
+			}
+		} else {
+			delete(cluster.Status.NodePools, np.Name)
+		}
```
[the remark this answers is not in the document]
wallrj: The whole function has been simplified and merged in a separate PR #299

pkg/controllers/cassandra/nodepool/nodepool.go
wallrj: This version of the function was just wrong. 
Instead, in #299  I re-populate the ``CassandraCluster.Status.NodePools`` each time, based on the statefulsets that are owned by the cluster, and which have the expected name.

pkg/controllers/cassandra/testing/gen.go
wallrj: the initial value is between 0-5 (see code above) so it should never be more than 6.
This is a way to simulate an occasional difference between the Spec.NodePool.Replicas and Status.NodePools.ReadyReplicas

internal/test/util/generate/generate.go
```diff
@@ -118,3 +120,46 @@ func StatefulSet(c StatefulSetConfig) *apps.StatefulSet {
 		},
 	}
 }
+
+func AssertStatefulSetMatches(t *testing.T, expected StatefulSetConfig, actual *apps.StatefulSet) {
```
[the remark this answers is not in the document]
wallrj: Removed that assertion and simplified the tests.

PR #258, opened 2018-02-26
Remove cassandra chart

kragniz:
We should encourage users to directly use the CassandraCluster and ElasticsearchCluster resources. The helm chart creates a layer of abstraction I don't think we need, and there is no corresponding elasticsearch chart.

This instead uses the same templating elasticsearch uses for e2e testing (envsubst).

/cc @mattbates 

```release-note
Remove cassandra chart
```


hack/testdata/cass-cluster-test.template.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: "navigator.jetstack.io/v1alpha1"
+kind: "CassandraCluster"
+metadata:
+  name: "test"
```
wallrj: This should be a CASS_NAME parameter I think

kragniz: :+1:

kragniz:
Now rebased on master with tests passing again

Issue #262, opened 2018-02-27
Cassandra Pilot should report the Cassandra version running on each node


wallrj:
In order to perform Cassandra upgrades we need to know the Cassandra version that is actually running on each node.

We can't simply look at the Docker image tag, because it may not follow semver versioning.

Instead the pilot should query Cassandra directly, either using CQL or via the JMX interface.

It should then record the discovered version on the `Pilot.Status` or perhaps on `CassandraCluster.Status.Pilots[<pilot_name>].Version`

With this information, the Cassandra controller can calculate when an `UpgradeCassandra` action is required.

/kind feature

PR #244, opened 2018-02-07
Add a CassandraCluster.Version field

wallrj:
* CassandraCluster now has a Version field which must be a valid semver.
* CassandraCluster.Image is now optional
* and if ommitted will default to the image from docker.io/cassandra matching the supplied semver version.
* If Image is supplied, its repository and tag must be supplied, but pullPolicy is optional.
* Created shared test cases for NavigatorClusterConfig validation.
* In the Cassandra helm chart, I've removed the ability to override the cassandra image tag and repository for now, because its not needed in any of the tests.

Fixes: #234

**Release note**:
```release-note
NONE
```

PR #263, opened 2018-02-27
Report the actual Cassandra Version

wallrj:
* Add a Pilot.Status.Cassandra.Version field.
* Add a nodetool.Version method which queries Cassandra version via JMX / Jolokia.
* A CassandraVersion type that wraps semver.Version in order to add a special case for versions without a patch number.
* Update the Pilot.Status.Cassandra.Version field during calls to Pilot.Sync.
* An E2E test for version reporting.

Fixes: #262 

**Release note**:
```release-note
NONE
```


wallrj:
Rebased after closing #244

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -299,6 +301,11 @@ type ElasticsearchPilotStatus struct {
 	Version semver.Version `json:"version,omitempty"`
 }
 
+type CassandraPilotStatus struct {
+	// Version as reported by the Elasticsearch process
```
kragniz: Elasticsearch -> Cassandra

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -80,9 +80,24 @@ func (p *Pilot) CmdFunc(pilot *v1alpha1.Pilot) (*exec.Cmd, error) {
...
+	pilot = pilot.DeepCopy()
+	if pilot.Status.Cassandra == nil {
+		pilot.Status.Cassandra = &v1alpha1.CassandraPilotStatus{}
+	}
+	version, err := p.nodeTool.Version()
+	if err != nil {
+		return err
+	}
+	pilot.Status.Cassandra.Version = version
+	_, err = p.navigatorClient.Navigator().Pilots(pilot.Namespace).UpdateStatus(pilot)
```
kragniz: Should we only perform the update if the version differs?

wallrj:
* Only update if version changed
* And if nodetool.Version fails, attempt to set the .Status.Version to nil
* Added some unit tests for Pilot.Sync

hack/e2e.sh
```diff
@@ -230,6 +230,16 @@ function test_cassandracluster() {
         fail_test "First cassandra node not marked as seed"
     fi
 
+    if ! retry TIMEOUT=300 \
+         stdout_equals "${CASS_VERSION}" \
+         kubectl --namespace "${namespace}" \
+         get pilots \
+         --output 'jsonpath={.items[*].status.cassandra.version}'
```
munnerz: '*' here will select versions on all pilots, no? How does this work if there are 2 pilots returned? (or perhaps there's not 2 pilots at the point this test is run?)

pkg/apis/navigator/types.go
```diff
@@ -206,6 +208,12 @@ type ElasticsearchPilotStatus struct {
 	Version   semver.Version
 }
 
+type CassandraPilotStatus struct {
+	// The version reported by the Cassandra process.
+	// nil indicates that the version is not yet known / not yet reported.
+	Version *semver.Version
```
munnerz: This comment is best on the v1alpha1 type (especially the extra info about nil). We won't be generating docs from internal types, only the external versions.

pkg/cassandra/version/version.go
```diff
@@ -0,0 +1,45 @@
...
+	s, err := strconv.Unquote(string(data))
+	if err != nil {
+		return err
+	}
+	err1 := v.Version.Set(s)
+	if err1 != nil {
+		err2 := v.Version.Set(s + ".0")
+		if err2 != nil {
+			return err1
+		}
```
munnerz: This is a very confusing dance of `err` vars. Can we flatten this to use `err` as much as possible, and if it isn't possible, give the second error a more helpful name than `err1`?

pkg/cassandra/version/version.go
```diff
@@ -0,0 +1,45 @@
...
+	if v.Version == nil {
+		v.Version = &semver.Version{}
+	}
+	s, err := strconv.Unquote(string(data))
+	if err != nil {
+		return err
+	}
+	err1 := v.Version.Set(s)
+	if err1 != nil {
+		err2 := v.Version.Set(s + ".0")
```
munnerz: I think we need a more thorough check of the existing version string before assuming we can just add `.0` onto the end. This also wouldn't handle something like `v3.0-alpha.1` correctly (as it'd be parsed as `v3.0-alpha.1.0`.

pkg/cassandra/version/version.go
```diff
@@ -0,0 +1,45 @@
...
+		if err2 != nil {
+			return err1
+		}
+	}
+	return nil
+}
+
+var _ json.Unmarshaler = &Version{}
```
munnerz: We also need to implement a marshaller that ensures that unmarshalling and then remarshalling is idempotent imo. This could cause confusion for users otherwise.

OTOH, perhaps we should actually disallow this invalid semver within Cassandra and *always* require valid semvers on our resources. If Cassandra reports an invalid semver, we could 'repair' it within the Pilot itself. Not sure of the implications of this, but given Cassandra is semver-like, it might work well & also saves our idempotentcy (I invented a new word) issues.

pkg/cassandra/nodetool/fake/fake.go
```diff
@@ -0,0 +1,40 @@
+package fake
+
+import (
+	"fmt"
+
+	"github.com/coreos/go-semver/semver"
```
munnerz: Should we be using our internal cassandra version type here?

pkg/cassandra/nodetool/nodetool_test.go
```diff
@@ -356,3 +358,26 @@ func TestNodeToolStatus(t *testing.T) {
...
+
+var _ client.Interface = &fakeClient{}
+
+func TestNodeToolVersion(t *testing.T) {
+	nt := nodetool.New(&fakeClient{})
+	ntResponse, ntErr := nt.Version()
+	if ntErr != nil {
+		t.Error(ntErr)
+	}
+	t.Log(ntResponse)
```
munnerz: We should probably verify this version is as expected, and not just that the function doesn't error. Also should we have a test for a case where the `Version` function *does* error?

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -80,9 +82,39 @@ func (p *Pilot) CmdFunc(pilot *v1alpha1.Pilot) (*exec.Cmd, error) {
...
 func (p *Pilot) syncFunc(pilot *v1alpha1.Pilot) error {
-	if p.genericPilot.IsThisPilot(pilot) {
+	if !p.genericPilot.IsThisPilot(pilot) {
+		return nil
 	}
-	return nil
+	pilot, err := p.pilotLister.Pilots(pilot.Namespace).Get(pilot.Name)
+	if err != nil {
+		return err
+	}
```
munnerz: Don't need to obtain a copy of `pilot` - you already have one as the arg to the sync func which was just retrieved from the lister (in the callsite for this function)

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -80,9 +82,39 @@ func (p *Pilot) CmdFunc(pilot *v1alpha1.Pilot) (*exec.Cmd, error) {
...
+	if err == nil {
+		lastReportedVersion := pilotCopy.Status.Cassandra.Version
+		if lastReportedVersion == nil || !lastReportedVersion.Equal(*version) {
+			pilotCopy.Status.Cassandra.Version = version
+		}
+	} else {
+		errs = append(errs, err)
+		pilotCopy.Status.Cassandra.Version = nil
+	}
+	if !reflect.DeepEqual(pilot, pilotCopy) {
```
munnerz: We should avoid DeepEqual as much as possible - it's slow and expensive. A simple bool for needsUpdate would suffice, especially as this block is quite simple and small at this point.

pkg/pilot/cassandra/v3/pilot_test.go
```diff
@@ -0,0 +1,147 @@
...
+			},
+		},
+		"nodetool version failure causes nil status version": {
+			navObjects: []runtime.Object{localPilotV300},
+			pilot:      localPilotV300,
+			nodetool:   fakenodetool.New().SetVersionError("simulated nodetool error"),
+			assertions: func(t *testing.T, state *controllers.State) {
+				CheckPilotVersion(
+					localPilot.Name, localPilot.Namespace, nil, t, state,
+				)
```
munnerz: I like this `assertions` field 😄

hack/e2e.sh
wallrj: There's only one pilot at this point, but you're right it'd be better give index-0

pkg/cassandra/version/version.go
wallrj: I've switched to using Cassandra version throughout.
And added the MarshalJSON method.

wallrj:
Thanks for the review @munnerz 

I've updated the patch to use the Cassandra `version.Version` through out....including in the Navigator API.
If a user specifies `CassandraCluster.Spec.Version: 3.9`  , 
 * the default image will be 3.9.0.
 * Pilot.Status.Cassandra.version will be 3.9

I'll add comparison methods in a followup branch, when I implement the Cassandra Upgrade action.

I think I've addressed all your other comments.

Please take another look it the E2E tests pass.

wallrj:
I think those test failures were caused by me using version: 3.11 in the E2E tests....which got pulled as 3.11.0 and failed without any debug logs.

I've added a `trap dump_debug_logs _artifacts EXIT` to ensure the logs get dumped, even if the script exits because of a `set -e` error....or indeed when the script exits without a failure.

Which should help debug similar test failures in future.

wallrj:
While working on the Upgrade action branch I kept seeing `pilot.status.cassandra` and `pilot.status.cassandra.version` flickering....sometimes it'd be there and next second it'd be nil.

I think the problem was that I was performing my own pilot status update...not realising that the generic pilot also does a status update.

pkg/cassandra/version/version.go
```diff
@@ -0,0 +1,110 @@
...
+// It is returned as a string via:
+// StorageService.getReleaseVersion: https://github.com/apache/cassandra/blob/cassandra-3.11.2/src/java/org/apache/cassandra/service/StorageService.java#L2863
+// FBUtilities.getReleaseVersionString: https://github.com/apache/cassandra/blob/cassandra-3.11.2/src/java/org/apache/cassandra/utils/FBUtilities.java#L326
+// Which appears to read the version string from a `Properties` API which appears to be created via this XML file: https://github.com/apache/cassandra/blob/cassandra-3.11.2/build.xml#L790
+// Internally, Cassandra converts the version string to a `CassandraVersion` object which supports rich comparison.
+// See https://github.com/apache/cassandra/blob/cassandra-3.11.2/src/java/org/apache/cassandra/utils/CassandraVersion.java
+// In Navigator we parse the Cassandra version string as early as possible, into a similar Cassandra Version object.
+// This also fixes the missing Patch number and stores the version internally as a semver.
+// It also keeps a reference to the original version string so that we can report that in our API.
+// So that the version reported in our API matches the version that an administrator expects.
```
kragniz: +1 good comments for context

PR #267, opened 2018-02-28
Add doc on multi-az cassandra deployments

kragniz:
```release-note
NONE
```

### DIFF
--- a/docs/cassandra/multi-az.md
+++ b/docs/cassandra/multi-az.md
@@ -1,6 +1,9 @@
 Cassandra Across Multiple Availability Zones
 ============================================
 
+With rack awareness
+-------------------
+
 Navigator supports running Cassandra with
 [rack and datacenter-aware replication](https://docs.datastax.com/en/cassandra/latest/cassandra/architecture/archDataDistributeReplication.html).
 To deploy this, you must run a `nodePool` in each availability zone, and mark each as a separate Cassandra rack.
@@ -51,4 +54,34 @@ As an example, the nodePool section of a CassandraCluster spec for deploying int
       enabled: true
       size: "5Gi"
       storageClass: "default"
+```
+
+Without rack awareness
+----------------------
+
+Since the default rack name is equal to the nodepool name,
+simply set the rack name to the same static value in each nodepool to disable rack awareness.
+
+A simplified example:
+
+```yaml
+  nodePools:
+  - name: "np-europe-west1-b"
+    replicas: 3
+    datacenter: "europe-west1"
+    rack: "default-rack"
+    nodeSelector:
+      failure-domain.beta.kubernetes.io/zone: "europe-west1-b"
+  - name: "np-europe-west1-c"
+    replicas: 3
+    datacenter: "europe-west1"
+    rack: "default-rack"
+    nodeSelector:
+      failure-domain.beta.kubernetes.io/zone: "europe-west1-c"
+  - name: "np-europe-west1-d"
+    replicas: 3
+    datacenter: "europe-west1"
+    rack: "default-rack"
+    nodeSelector:
+      failure-domain.beta.kubernetes.io/zone: "europe-west1-d"
 ```
--- a/docs/cassandra/multi-az.md
+++ b/docs/cassandra/multi-az.md
@@ -1,20 +1,16 @@
-Cassandra across multiple availability zones
+Cassandra Across Multiple Availability Zones
 ============================================
 
-Navigator supports running Cassandra with rack and datacenter-aware
-replication. To deploy this, you must run a `nodePool` in each availability
-zone, and mark each as a separate Cassandra rack.
+Navigator supports running Cassandra with rack and datacenter-aware replication.
+To deploy this, you must run a `nodePool` in each availability zone, and mark each as a separate Cassandra rack.
 
-The `nodeSelector` field of a nodePool allows scheduling the nodePool to a set
-of nodes matching labels. This should be used with a node label such as
-`failure-domain.beta.kubernetes.io/zone`.
+The `nodeSelector` field of a nodePool allows scheduling the nodePool to a set of nodes matching labels.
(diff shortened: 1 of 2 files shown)


docs/cassandra/multi-az.md
```diff
@@ -0,0 +1,51 @@
+Cassandra across multiple availability zones
```
wallrj: Maybe Use Title Case

docs/cassandra/multi-az.md
```diff
@@ -0,0 +1,51 @@
...
+============================================
+
+Navigator supports running Cassandra with rack and datacenter-aware
+replication. To deploy this, you must run a `nodePool` in each availability
+zone, and mark each as a separate Cassandra rack.
```
wallrj: Add a link to Cassandra docs on DC / rack awareness.

docs/cassandra/multi-az.md
```diff
@@ -0,0 +1,51 @@
...
+
+The `nodeSelector` field of a nodePool allows scheduling the nodePool to a set
+of nodes matching labels. This should be used with a node label such as
+`failure-domain.beta.kubernetes.io/zone`.
```
wallrj: :grey_question: Maybe add a link to kubernetes failure-domain label docs. If there are any.

docs/cassandra/multi-az.md
```diff
@@ -0,0 +1,51 @@
...
+
+The `datacenter` and `rack` fields mark all Cassandra nodes in a nodepool as
+being located in that datacenter and rack. This information can then be used
+with the `NetworkTopologyStrategy` keyspace replica placement strategy.
```
wallrj: :grey_question: Link to NetworkTopologyStrategy docs.

docs/cassandra/multi-az.md
```diff
@@ -0,0 +1,51 @@
...
+
+As an example, the nodePool section of a CassandraCluster spec for deploying
+into GKE in europe-west1:
+
+```yaml
+  nodePools:
+  - name: "np-europe-west1-b"
+    replicas: 3
+    datacenter: "europe-west1"
+    rack: "europe-west1-b"
```
wallrj: :grey_question: This makes me wonder what happens if I used `rack: "b"`. Does Cassandra know that rack b in europe-west1 is different than rack b in us-west1?

wallrj: And if not, perhaps we should add a note that rack labels must be unique.

kragniz: Ec2Snitch simply labels a rack in `us-east-1a` as `1a`, so it seems cassandra knows the difference:

https://github.com/apache/cassandra/blob/4a50f4467f439f9a00aac6aa4d3b83a6c1c8d6c4/src/java/org/apache/cassandra/locator/Ec2Snitch.java#L54-L62

kragniz:
The current hack/verify-links.sh is failing to parse the markdown added in this patch, so I've updated it with the version in https://github.com/duglin/vlinker which seems to work better.

docs/cassandra/multi-az.md
```diff
@@ -0,0 +1,51 @@
+Cassandra Across Multiple Availability Zones
+============================================
+
+Navigator supports running Cassandra with
+[rack and datacenter-aware replication](https://docs.datastax.com/en/cassandra/latest/cassandra/architecture/archDataDistributeReplication.html).
+To deploy this, you must run a `nodePool` in each availability zone, and mark each as a separate Cassandra rack.
+
+The `nodeSelector` field of a nodePool allows scheduling the nodePool to a set of nodes matching labels.
+This should be used with a node label such as
+[`failure-domain.beta.kubernetes.io/zone`](https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesiozone).
```
munnerz: Would be good to add a link out to the official docs on nodeSelector. It's not a concept we've invented and there are probably better descriptions out there that we want to include :smile:

docs/cassandra/multi-az.md
```diff
@@ -0,0 +1,51 @@
...
+
+The `datacenter` and `rack` fields mark all Cassandra nodes in a nodepool as being located in that datacenter and rack.
+This information can then be used with the
+[`NetworkTopologyStrategy`](http://cassandra.apache.org/doc/latest/architecture/dynamo.html#network-topology-strategy)
+keyspace replica placement strategy.
+
+As an example, the nodePool section of a CassandraCluster spec for deploying into GKE in europe-west1:
```
munnerz: ... "with rack awareness/network topology strategy/whatever it is called enabled"

docs/cassandra/multi-az.md
```diff
@@ -0,0 +1,51 @@
...
+    persistence:
+      enabled: true
+      size: "5Gi"
+      storageClass: "default"
+  - name: "np-europe-west1-c"
+    replicas: 3
+    datacenter: "europe-west1"
+    rack: "europe-west1-c"
+    nodeSelector:
+      failure-domain.beta.kubernetes.io/zone: "europe-west1-c"
```
munnerz: Did we drop the "auto-pull zone from nodeSelector into rack attribute" functionality for now?

kragniz: Yes, I figure it's better to be explicit about it for now. We can add it in later if we think it'll be useful.

docs/cassandra/multi-az.md
kragniz: :+1:

munnerz:
As we default the 'rack' to the node pool name, perhaps we should also document how someone would **disable** rack awareness? I worry that we've made it harder for users to do this, and it might be unexpected during some failure scenario (and difficult to change after initial deployment). Otherwise lgtm - happy to add the lgtm label once you've responded to this query.

kragniz:
I updated with two sections - with/without rack awareness

PR #269, opened 2018-02-28
Move rack and datacenter defaulting to API

kragniz:
This allows the defaults to be visable in something like `kubectl get
cassandraclusters`, avoiding the user being surprised.

```release-note
NONE
```

### DIFF
--- a/pkg/apis/navigator/v1alpha1/defaults.go
+++ b/pkg/apis/navigator/v1alpha1/defaults.go
@@ -4,6 +4,23 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const (
+	cassDefaultDatacenter = "navigator-default-datacenter"
+)
+
 func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
+}
+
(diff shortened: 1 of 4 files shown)


munnerz:
This should be moved into pkg/apis/navigator/v1alpha1 as an actual
defaulting function and not as part of the apiserver registry directly.

See an example here:
https://github.com/kubernetes-incubator/service-catalog/blob/master/pkg/apis/servicecatalog/v1beta1/defaults.go

kragniz:
@munnerz  ah nice, I didn't know about the defaulter-gen stuff

pkg/apis/navigator/v1alpha1/defaults.go
```diff
@@ -4,6 +4,23 @@ import (
...
 
+const (
+	cassDefaultDatacenter = "navigator-default-datacenter"
+)
+
 func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
+
+func SetDefaults_CassandraClusterSpec(spec *CassandraClusterSpec) {
```
kragniz: I'll move this to a `SetDefaults_CassandraClusterNodePool` function as well

PR #270, opened 2018-02-28
Move seed pod labelling to a separate controller

wallrj:
* In #256 I want to refactor nodepool.Sync so that its only responsibility is to update the NodePool status.
* And the seed labelling seems like it is a separate concern that can live in its own module.
* With its own unit tests.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -222,13 +222,6 @@ function test_cassandracluster() {
         fail_test "Cassandra pilots did not elect a leader"
     fi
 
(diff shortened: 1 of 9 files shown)


wallrj:
Test failures were caused by single double quoted map keys in the jsonpath expression.
I've made them single quoted now.
With an idea for a future improvement to the E2E tests for seed node service.

wallrj:
Try again, without the funky shell quoting.

wallrj:
Use a label selector instead.
And do the test after scaling out the cluster.

PR #271, opened 2018-02-28
WIP: Add script to update example manifests

### LINKED ISSUE #260: Sync cluster templates used in e2e tests with examples in docs
/kind feature

Currently it's easy for the example clusters to get out of date. Instead of maintaining two separate definitions, we should have a script that generates the examples based on the templates used in e2e tests. This ensures the examples are valid (since they've passed e2e tests), and reduces the amount of effort to maintain them.

We could additionally add a check in CI to make sure the examples are kept in sync.

kragniz:
Fixes #260.

```release-note
NONE
```

### DIFF
--- a/hack/testdata/cass-cluster-test.template.yaml
+++ b/hack/testdata/cass-cluster-test.template.yaml
@@ -11,7 +11,7 @@ spec:
   - name: "ringnodes"
     replicas: ${CASS_REPLICAS}
     datacenter: "${CASS_NAME}-datacenter"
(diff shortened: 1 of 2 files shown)


munnerz:
We also need a corresponding `verify` script here.

FWIW, as I've said before, I **do not think** our e2e tests should only be driven by docs. Instead, we should have tests for the things in our docs (i.e. using sphinx's testing features), but the e2e test suite should test more advanced features which may or may not make sense to include in our documentation.

PR #276, opened 2018-03-06
Configurable scheduler

### LINKED ISSUE #235: Support scheduler specification in cluster node pool specification
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**:

Pods are placed on any node satisfying nodeSelector.

**What you expected to happen**:

Pods are placed nodes that the persistent volume is provisioned from

**How to reproduce it (as minimally and precisely as possible)**:

Run Portworx for persistent volume provider. Install [stork](https://github.com/libopenstorage/stork).

**Anything else we need to know?**:

To place pods on nodes with persistent volume, the scheduler needs to be configurable so the name given to the stork scheduler can be used.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.8.4_coreos.0
- Cloud provider or hardware configuration**: Azure L8s VMs
- Install tools:
- Others:

kragniz:
Fixes #235

This allows users to configure the scheduler for pods created via navigator's statefulsets. It's a fairly simple passthrough of the value to the statefulset spec.

There are two options exposed, one in the cluster spec, and the other in the nodepool spec. The nodepool value takes precedent, falling back to the cluster value if it is defined, then the default scheduler.

For example, these (simplified) are equivalent:
```yaml
kind: "CassandraCluster"              
metadata:                             
  name: "example"                    
spec:                                 
  nodePools:                          
  - name: "ringnodes-1"        
    schedulerName: "fancy-scheduler"
  - name: "ringnodes-2"        
    schedulerName: "fancy-scheduler"
```

```yaml
kind: "CassandraCluster"              
metadata:                             
  name: "example"                    
spec:                                 
  schedulerName: "fancy-scheduler"              
  nodePools:                          
  - name: "ringnodes-1"        
  - name: "ringnodes-2"        
```

```release-note
Allow configuring scheduler
```

### DIFF
--- a/docs/configure-scheduler.rst
+++ b/docs/configure-scheduler.rst
@@ -1,17 +1,9 @@
-Configure scheduler type
+Configure Scheduler Type
 ------------------------
 
-If a custom scheduler type is required (for example if you are deploying with Portworx or another storage provider), this can be set on either each nodepool:
-
-.. code-block:: yaml
-
-    spec:
-      schedulerName: "fancy-scheduler"
-      nodePools:
-      - name: "ringnodes-1"
-      - name: "ringnodes-2"
-
-or for an entire cluster:
+If a `custom scheduler <https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/>`_ type is required
+(for example if you are deploying with `stork <https://docs.portworx.com/scheduler/kubernetes/stork.html>`_ or another storage provider),
+this can be set on each nodepool:
 
 .. code-block:: yaml
 
@@ -22,4 +14,4 @@ or for an entire cluster:
       - name: "ringnodes-2"
         schedulerName: "fancy-scheduler"
 
-The nodepool field takes precedent, falling back to the cluster value if it is defined, then the default scheduler.
+If the nodepool field is not specified, the default scheduler is used.
--- a/docs/configure-scheduler.rst
+++ b/docs/configure-scheduler.rst
@@ -1,0 +1,25 @@
+Configure scheduler type
+------------------------
+
+If a custom scheduler type is required (for example if you are deploying with Portworx or another storage provider), this can be set on either each nodepool:
+
+.. code-block:: yaml
+
+    spec:
+      schedulerName: "fancy-scheduler"
+      nodePools:
+      - name: "ringnodes-1"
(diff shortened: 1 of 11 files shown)


docs/configure-scheduler.rst
```diff
@@ -0,0 +1,25 @@
+Configure scheduler type
+------------------------
```
wallrj: Title case

docs/configure-scheduler.rst
```diff
@@ -0,0 +1,25 @@
...
+
+If a custom scheduler type is required (for example if you are deploying with Portworx or another storage provider), this can be set on either each nodepool:
```
wallrj: Doesn't sound quite right. 
Maybe instead "can be set on either each nodepool" > "can either be set on the cluster and / or on each nodepool"

docs/configure-scheduler.rst
```diff
@@ -0,0 +1,25 @@
...
+.. code-block:: yaml
+
+    spec:
+      nodePools:
+      - name: "ringnodes-1"
+        schedulerName: "fancy-scheduler"
+      - name: "ringnodes-2"
+        schedulerName: "fancy-scheduler"
+
+The nodepool field takes precedent, falling back to the cluster value if it is defined, then the default scheduler.
```
wallrj: Typo (maybe) "takes precedence"

wallrj: Maybe link to the Kubernetes custom schedulers documentation or the portworx docs:
 * E.g. See [Configure Multiple Schedulers](https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/) for more information about Kubernetes with multiple schedulers.

munnerz:
> I'm not convinced of the benefit of allowing default schedulername on the cluster. What if you change the default schedulername and then add another nodepool. I'd expect the current nodepools to also adopt the new default, but they'll be left with the original default.

This is a good point - we need to be careful with how we validate changes if we have this top level field. Perhaps we can remove it for now, and in future add it if required?

kragniz:
Yep, makes sense to me. I'll remove in the next patchset

kragniz:
I've made the update validation for nodePools a whitelist of changes in #283, so with those patches SchedulerName will not be mutable.

Issue #232, opened 2018-02-01
Document and demonstrate how applications should connect to Cassandra


wallrj:
I added a cassandra CQL service which loadbalances connections to the CQL ports of all Cassandra nodes.

But on further investigation, it seems that clients (such as GoCQL) will discover all the C* nodes automatically, given a connection to one of them.
And the loadbalancer IP of the CQL service might even confuse this process, since it's not an IP address advertised by any of the nodes.

Figure out the best practice for connecting to Cassandra and document it and perhaps add an example to the documentation and to the E2E tests.

 * https://godoc.org/github.com/gocql/gocql#NewCluster

Get rid of the CQL service if it is found to be unnecessary.

/kind feature

PR #278, opened 2018-03-08
Remove the loadbalanced CQL service and document an alternative way to connect.

wallrj:
In no particular order:

* Removed the CQL service.
* Added a headless `nodes` service.
* Renamed the `seedprovider` service to `seeds`.
* Added documentation explaining the two headless services and how to connect a CQL client.
* Refactored the two headless services into a single service control
* Removed the service update code, which was not tested and which added unecessary complication.
* Navigator will now only touch the services if they are missing.
* Updated the E2E tests so that `cql_connect` always attempts to connect to the `nodes` service.
* Removed the ServiceName from the NodePool statefulsets because it no longer made sense with multiple NodePools / StatefulSets. The SS servicename is supposed to match a service dedicated to that statefulset, not a single seedprovider service. We probably should dynamically create a service for each nodepool.
* Removed the, also broken, CASSANDRA_SEEDS configuration which was pointing `seedProviderServiceName` rather than at a service name matching the name of the statefulset.
* In E2E tests Reverted to a better mechanism for simulating node failure. Decommission leaves the node in a decommissioned state causing the C* process to immediately exit.
* Added CONSISTENCY checks to the E2E CQL queries to verify that both C* nodes are reachable and have the test data.

Fixes: #232

**Release note**:
```release-note
NONE
```


munnerz:
> Removed the, also broken, CASSANDRA_SEEDS configuration which was pointing seedProviderServiceName rather than at a service name matching the name of the statefulset.

Perhaps I misunderstand this bullet - when a node bootstraps, surely it should (attempt to) connect to *all* seeds in the cluster, not just its own node pool, in order to start?

wallrj:
In https://github.com/jetstack/navigator/pull/278#issuecomment-372641959 @munnerz wrote:
> Perhaps I misunderstand this bullet - when a node bootstraps, surely it should (attempt to) connect to all seeds in the cluster, not just its own node pool, in order to start?

There are two things:
 * As far as we can see, this environment variable is obsolete, now that we're using the Kubernetes seed provider. The CASSANDRA_SEEDS env variable results in a Cassandra configuration stanza which is specifically for the `SimpleSeedProvider`.
 * The `StatefulSet.Spec.ServiceName` had previously been set to a single, cluster wide service name. So when there are multiple node pools and pods from multiple statefulsets, the DNS names no longer reflect the controlling statefulset...instead they'd be something like  `euwest1a-0.cass-seeds.my-namespace.srv.cluster.local` which works (somehow)....but
* The domain name label `cass-seeds` is no longer ideal, because there will only be DNS names for pods that have been labelled as seeds. 
* Instead we want stable DNS names for all the pods in all statefulsets, not just for the seed pods.

I think we should follow this up with a branch that enhances `CreateNodePool` action so that it also creates a headless service per nodepool....and that dynamically created service can be the name we use in the `statefulset.spec.serviceName`

docs/cassandra.rst
```diff
@@ -116,3 +116,64 @@ The ``resources`` field follows exactly the same specification as the Kubernetes
...
+
+.. code-block:: python
+
+   from cassandra.cluster import Cluster
+
+   cluster = Cluster(['cass-demo-nodes'], port=9042)
+   session = cluster.connect()
+   rows = session.execute('SELECT ... FROM ...')
+   for row in rows:
+       print row
```
kragniz: not python3 :o

PR #279, opened 2018-03-12
Add shortName for cassandracluster

kragniz:
So commands like `kubectl get cass` are shorter.

```release-note
none
```

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -209,8 +209,8 @@ function test_cassandracluster() {
 
     if ! kubectl get \
          --namespace "${namespace}" \
-         cass; then
-         fail_test "Failed to get cassandraclusters short name (cass)"
+         cassandra; then
+         fail_test "Failed to get cassandraclusters short name (cassandra)"
     fi
 
     if ! kubectl apply \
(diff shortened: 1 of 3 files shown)


wallrj:
I've got mixed feelings about this.
* I'm not a fan of short names....I think they make shell scripts unclear, and kubectl command line completion takes away the pain of typing out the full name.
* Perhaps I'd suggest `kubectl get cassandra`
* But happy to go with this if others are in favour.
* What ever we choose let's use both short and long forms in the E2E tests, so that we can see it working.
* A simple `kubectl get cass` at the start of the test would be sufficient.

munnerz:
`kubectl get cassandra` feels more natural and sets a precedent for how we shorten names in future (e.g. `CouchbaseCluster` -> `couchbase`) etc.

I'm not bothered about `cass` too much, but it does seems to be an arbitrarily shortened version of `cassandra` and isn't particularly obvious. wdyt @kragniz?

kragniz:
`cass` was pretty random, I'm happy with `cassandra` instead.

munnerz:
ping @kragniz

kragniz:
Updated with `cassandra` and `elasticsearch` shortnames

PR #283, opened 2018-03-13
Make nodepool persistence settings immutable

### LINKED ISSUE #233: Changing CassandraCluster persistence settings should not be allowed
~~Right now, if you change the persistence settings for a running cluster, the navigator controller will attempt to update the statefulset with the new volumeclaimtemplate settings.~~

~~I'm not sure whether this is supported and even if so, Navigator doesn't yet have a way to control the rollout of the updated statefulset.~~

We should prevent the user from changing the persistence setting for now.

See:
 * https://github.com/kubernetes/kubernetes/issues/41015
 * https://github.com/kubernetes/kubernetes/pull/46669
 * But https://github.com/kubernetes/community/pull/503 suggests that updating storage settings was out of scope.
 * https://github.com/kubernetes/kubernetes/issues/35941

/kind bug

kragniz:
**What this PR does / why we need it**: this adds validation to not allow changing persistence settings in cassandra and elasticsearch, and adds test coverage for the update validation.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #233

**Release note**:

```release-note
none
```

### DIFF
--- a/pkg/apis/navigator/validation/cassandra.go
+++ b/pkg/apis/navigator/validation/cassandra.go
@@ -33,19 +33,23 @@ func ValidateCassandraClusterUpdate(old, new *navigator.CassandraCluster) field.
 
 		for _, oldNp := range old.Spec.NodePools {
 			if newNp.Name == oldNp.Name {
-				if newNp.Rack != oldNp.Rack {
-					allErrs = append(allErrs, field.Forbidden(idxPath.Child("rack"), "cannot modify rack"))
-				}
-				if newNp.Datacenter != oldNp.Datacenter {
-					allErrs = append(allErrs, field.Forbidden(idxPath.Child("datacenter"), "cannot modify datacenter"))
-				}
-				if !reflect.DeepEqual(newNp.NodeSelector, oldNp.NodeSelector) {
-					allErrs = append(allErrs, field.Forbidden(idxPath.Child("nodeSelector"), "cannot modify nodeSelector"))
+				if !reflect.DeepEqual(newNp.Persistence, oldNp.Persistence) {
+					if oldNp.Persistence.Enabled {
+						allErrs = append(allErrs, field.Forbidden(idxPath.Child("persistence"), "cannot modify persistence configuration once enabled"))
+					}
 				}
 
-				if !reflect.DeepEqual(newNp.Persistence, oldNp.Persistence) {
-					allErrs = append(allErrs, field.Forbidden(idxPath.Child("persistence"), "cannot modify persistence configuration"))
+				restoreReplicas := newNp.Replicas
+				newNp.Replicas = oldNp.Replicas
+
+				restorePersistence := newNp.Persistence
+				newNp.Persistence = oldNp.Persistence
+
+				if !reflect.DeepEqual(newNp, oldNp) {
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), "updates to nodepool for fields other than 'replicas' and 'persistence' are forbidden."))
 				}
+				newNp.Replicas = restoreReplicas
+				newNp.Persistence = restorePersistence
 
 				break
 			}
--- a/pkg/apis/navigator/validation/cassandra.go
+++ b/pkg/apis/navigator/validation/cassandra.go
@@ -42,6 +42,11 @@ func ValidateCassandraClusterUpdate(old, new *navigator.CassandraCluster) field.
 				if !reflect.DeepEqual(newNp.NodeSelector, oldNp.NodeSelector) {
 					allErrs = append(allErrs, field.Forbidden(idxPath.Child("nodeSelector"), "cannot modify nodeSelector"))
 				}
+
+				if !reflect.DeepEqual(newNp.Persistence, oldNp.Persistence) {
+					allErrs = append(allErrs, field.Forbidden(idxPath.Child("persistence"), "cannot modify persistence configuration"))
+				}
+
 				break
 			}
 		}
--- a/pkg/apis/navigator/validation/cassandra_test.go
+++ b/pkg/apis/navigator/validation/cassandra_test.go
@@ -20,6 +20,13 @@ var (
 			Version: *version.New("5.6.2"),
 			Image:   &validImageSpec,
 			NavigatorClusterConfig: validNavigatorClusterConfig,
+			NodePools: []navigator.CassandraClusterNodePool{
+				navigator.CassandraClusterNodePool{
+					Datacenter:  "datacenter-1",
+					Rack:        "rack-1",
+					Persistence: validNodePoolPersistenceConfig,
+				},
+			},
 		},
 	}
 )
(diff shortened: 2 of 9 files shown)


pkg/apis/navigator/validation/cassandra.go
```diff
@@ -40,6 +40,11 @@ func ValidateCassandraClusterUpdate(old, new *navigator.CassandraCluster) field.
 				if !reflect.DeepEqual(newNp.NodeSelector, oldNp.NodeSelector) {
 					allErrs = append(allErrs, field.Forbidden(idxPath.Child("nodeSelector"), "cannot modify nodeSelector"))
 				}
+
+				if !reflect.DeepEqual(newNp.Persistence, oldNp.Persistence) {
```
munnerz: If persistence was previously disabled, we probably can allow changes that enable the field later?

pkg/apis/navigator/validation/elasticsearch.go
```diff
@@ -102,6 +103,32 @@ func ValidateElasticsearchCluster(esc *navigator.ElasticsearchCluster) field.Err
...
+	fldPath := field.NewPath("spec")
+
+	npPath := fldPath.Child("nodePools")
+	for i, newNp := range new.Spec.NodePools {
+		idxPath := npPath.Index(i)
+
+		for _, oldNp := range old.Spec.NodePools {
+			if newNp.Name == oldNp.Name {
+				if !reflect.DeepEqual(newNp.NodeSelector, oldNp.NodeSelector) {
+					allErrs = append(allErrs, field.Forbidden(idxPath.Child("nodeSelector"), "cannot modify nodeSelector"))
```
munnerz: Why do we not allow modifying the nodeSelector? It's an updatable field in StatefulSet. If there are issues around how it plays with the `rack` parameter/node selector targeting a particular zone, we may need to rethink what we're doing there.

pkg/apis/navigator/validation/elasticsearch.go
```diff
@@ -102,6 +103,32 @@ func ValidateElasticsearchCluster(esc *navigator.ElasticsearchCluster) field.Err
...
+
+	fldPath := field.NewPath("spec")
+
+	npPath := fldPath.Child("nodePools")
+	for i, newNp := range new.Spec.NodePools {
+		idxPath := npPath.Index(i)
+
+		for _, oldNp := range old.Spec.NodePools {
+			if newNp.Name == oldNp.Name {
+				if !reflect.DeepEqual(newNp.NodeSelector, oldNp.NodeSelector) {
```
munnerz: Is reflect.DeepEqual usually used for Update validations? If so, 👍

kragniz: Yep, that's what's used upstream

pkg/apis/navigator/validation/cassandra.go
kragniz: I've allowed users to enable, but not disable

pkg/apis/navigator/validation/elasticsearch.go
kragniz: We'd need an associated action to manipulate the nodeSelector - currently we only modify replicas.

pkg/apis/navigator/validation/cassandra.go
```diff
@@ -33,15 +33,24 @@ func ValidateCassandraClusterUpdate(old, new *navigator.CassandraCluster) field.
...
-					allErrs = append(allErrs, field.Forbidden(idxPath.Child("datacenter"), "cannot modify datacenter"))
-				}
-				if !reflect.DeepEqual(newNp.NodeSelector, oldNp.NodeSelector) {
-					allErrs = append(allErrs, field.Forbidden(idxPath.Child("nodeSelector"), "cannot modify nodeSelector"))
+
+				restoreReplicas := newNp.Replicas
+				newNp.Replicas = oldNp.Replicas
+
+				restorePersistence := newNp.Persistence
+				newNp.Persistence = oldNp.Persistence
```
wallrj: :grey_question: I wonder why it's necessary to restore the original values? I see that's what they do in the StatefulSet validation, but I thought that validation didn't have opportunity to mutate the supplied objects. And I guess you can't mutate the old object either, otherwise why not do this the other way round? Assign the new value to the old object and then compare?

pkg/apis/navigator/validation/cassandra_test.go
```diff
@@ -86,3 +93,85 @@ func TestValidateCassandraCluster(t *testing.T) {
...
+				if !tc.errorExpected && len(errs) != 0 {
+					t.Errorf("unexpected errors: %s", errs)
+				}
+				for _, e := range errs {
+					t.Logf("error string is: %s", e)
+				}
+			},
+		)
+	}
+}
```
wallrj: :grey_question: tests look good to me, but it gets me thinking whether this might be a good place to use  property based tests (`testing/quick` or gopter). We'd generate a "range" of `CassandraCluster` objects and feed them to the validation function and then assert that if the supplied old and new objects had differences in anything but the whitelisted fields, that a validation error is generated. 
I tried something along these lines in https://github.com/jetstack/navigator/pull/256/commits/490933cab1f77881fce0227abe381a2f99de97d9#diff-37ffe475170e4d658b593d0134efd192R49 and I wouldn't mind discussing the idea sometime.

pkg/apis/navigator/validation/elasticsearch_test.go
```diff
@@ -472,3 +468,62 @@ func TestValidateElasticsearchClusterSpec(t *testing.T) {
...
+				if !tc.errorExpected && len(errs) != 0 {
+					t.Errorf("unexpected errors: %s", errs)
+				}
+				for _, e := range errs {
+					t.Logf("error string is: %s", e)
+				}
+			},
+		)
+	}
+}
```
wallrj: I ran the tests locally and couldn't see the equivalent for ElasticSearch of:

```
    --- PASS: TestValidateCassandraClusterUpdate/changed_rack (0.00s)
    	cassandra_test.go:172: error string is: spec: Forbidden: updates to nodepool for fields other than 'replicas' and 'persistence' are forbidden.
```

Do we need an extra test for changing something other than persistence on  `ElasticSearchNodePool`
Ignore me if you think this is nit-picking.

pkg/apis/navigator/validation/cassandra.go
kragniz: They do that in the statefulset validation because they do additional validation afterwards (`apivalidation.ValidateNonnegativeField`). I added it in case we want to do the same in the future, but can remove if it's extra noise.

pkg/apis/navigator/validation/elasticsearch_test.go
kragniz: Sure, I'll add extra test coverage

Issue #286, opened 2018-03-14
`CassandraCluster.Spec.Sysctl` does nothing


wallrj:
In the Cassandra controller I never added the init containers necessary to perform the sysctls.
And so far everything has been working.
So I think we should remove sysctls and document an alternative.

Seems like it should be something configured when provisioning the kubernetes node.
And then there should be a way of scheduling elasticsearch / cassandra pods to only the nodes where those sysctls have been configured.

Part of: #194 

/kind bug

PR #287, opened 2018-03-14
Remove the sysctl feature and add documentation instead

wallrj:
* Remove sysctl from example and test manifests.
* Remove sysctl from the API.
* Remove the sysctl init containers
* Remove the sysctl API validation code
* Add documentation about Elasticsearch OS configuration
* Add a daemonset to configure the E2E nodes with the required virtual memory settings (https://github.com/kubernetes/kubernetes/issues/36601 would be a better solution).

Fixes: #286 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/docs/quick-start/sysctl-daemonset.yaml
+++ b/docs/quick-start/sysctl-daemonset.yaml
@@ -1,0 +1,41 @@
+# Apply sysctl configuration required by Elasticsearch
+#
+# This DaemonSet will re-run sysctl every 60s on all nodes.
+#
+# XXX See CronJob daemonset which will allow scheduling one-shot or repeated
+# jobs across nodes:
+# https://github.com/kubernetes/kubernetes/issues/36601
+
+apiVersion: "extensions/v1beta1"
+kind: "DaemonSet"
+metadata:
+  name: "navigator-elasticsearch-sysctl"
+  namespace: "kube-system"
+spec:
+  template:
+    metadata:
+      labels:
+        app: "navigator-elasticsearch-sysctl"
+    spec:
+      containers:
+      - name: "apply-sysctl"
+        image: "busybox:latest"
+        resources:
+          limits:
+            cpu: "10m"
+            memory: "8Mi"
+          requests:
+            cpu: "10m"
+            memory: "8Mi"
+        securityContext:
+          privileged: true
+        command:
+        - "/bin/sh"
+        - "-c"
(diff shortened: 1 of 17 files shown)


hack/prepare-e2e.sh
```diff
@@ -32,6 +32,37 @@ items:
...
+              memory: 8Mi
+            requests:
+              cpu: 10m
+              memory: 8Mi
+          securityContext:
+            privileged: true
+          command:
+          - "/bin/sh"
+          - "-c"
+          - "sysctl -w vm.max_map_count=262144; sleep 2073600;"
```
munnerz: Would it be worth re-setting this more frequently, say, every 60s?

It's unlikely to ever be an issue, but if some other tool comes along and changes it, nodes may not start. Not a major blocker as this is obviously only used for our e2e tests, but if we want to let this serve as an example it may be best 😄

wallrj: I was actually looking for a way to sleep forever and discovered that  `sleep infintity` is a thing....but not on busy box.
Turns that that `sleep infinity` actually means `sleep 2073600` according to https://stackoverflow.com/a/41655546

That's why I chose it.

I propose I leave this for now and modify in a future branch?

munnerz:
We should also update the quick start guide to include a note about this: https://github.com/jetstack/navigator/blob/master/docs/quick-start.rst

docs/quick-start/sysctl-daemonset.yaml
```diff
@@ -0,0 +1,39 @@
...
+        securityContext:
+          privileged: true
+        command:
+        - "/bin/sh"
+        - "-c"
+        - |
+          set -o errexit
+          set -o xtrace
+          sysctl -w vm.max_map_count=262144
+          sleep 60s
```
munnerz: We need to actually loop in here when applying the value, else Kubernetes will eventually put the pod into a CrashLoopBackoff state

wallrj: Yeah, I thought about that, but wasn't sure if there'd be CrashLoopBackoff if the process exits with zero. I'll add the loop.

See:
 * https://github.com/kubernetes/kubernetes/issues/58619
 * https://github.com/kubernetes/kubernetes/issues/50962
 * https://github.com/kubernetes/kubernetes/issues/53125

PR #291, opened 2018-03-20
Use new test fixture in Cassandra controls

wallrj:
I'm trying to get rid of the Cassandra specific test fixture in `./pkg/controllers/cassandra/testing`
In this branch I've switched:
* ServiceAccounts
* Roles
* RoleBindings

I'll deal with Pilot and NodePool controls in separate branches.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cassandra/role/control.go
+++ b/pkg/controllers/cassandra/role/control.go
@@ -41,12 +41,16 @@ func NewControl(
 func (c *control) Sync(cluster *v1alpha1.CassandraCluster) error {
 	newRole := RoleForCluster(cluster)
 	client := c.kubeClient.RbacV1beta1().Roles(newRole.Namespace)
-	_, err := c.roleLister.
+	existingRole, err := c.roleLister.
 		Roles(newRole.Namespace).
 		Get(newRole.Name)
-	if k8sErrors.IsNotFound(err) {
-		_, err = client.Create(newRole)
+	if err == nil {
+		return util.OwnerCheck(existingRole, cluster)
 	}
+	if !k8sErrors.IsNotFound(err) {
+		return err
+	}
+	_, err = client.Create(newRole)
 	return err
 }
 
--- a/pkg/controllers/cassandra/role/control_test.go
+++ b/pkg/controllers/cassandra/role/control_test.go
@@ -1,40 +1,85 @@
 package role_test
 
 import (
-	"fmt"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/jetstack/navigator/internal/test/unit/framework"
+	"github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
+	"github.com/jetstack/navigator/pkg/controllers"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra/role"
 	casstesting "github.com/jetstack/navigator/pkg/controllers/cassandra/testing"
 )
 
 func TestRoleSync(t *testing.T) {
-	t.Run(
-		"role missing",
-		func(t *testing.T) {
-			f := casstesting.NewFixture(t)
-			f.Run()
-			f.AssertRolesLength(1)
+	cluster1 := casstesting.ClusterForTest()
+	role1 := role.RoleForCluster(cluster1)
+	foreignRole1 := role1.DeepCopy()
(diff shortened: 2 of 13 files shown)


pkg/controllers/cassandra/role/control.go
```diff
@@ -44,8 +44,15 @@ func (c *control) Sync(cluster *v1alpha1.CassandraCluster) error {
...
-		_, err = client.Create(newRole)
+	if err == nil {
+		return nil
+	}
+	if !k8sErrors.IsNotFound(err) {
+		return err
+	}
+	_, err = client.Create(newRole)
+	if k8sErrors.IsAlreadyExists(err) {
+		return nil
```
munnerz: We should probably return an error here still, as the resource may not/probably isn't like `newRole` (because it already exists). Then on the next sync, the `IsNotFound` check will be run again and the appropriate action taken.

pkg/controllers/cassandra/role/control_test.go
```diff
@@ -1,40 +1,93 @@
...
-			f := casstesting.NewFixture(t)
-			f.Run()
-			f.AssertRolesLength(1)
+	cluster1 := casstesting.ClusterForTest()
+	role1 := role.RoleForCluster(cluster1)
+
+	type testT struct {
+		kubeObjects        []runtime.Object
+		cluster            *v1alpha1.CassandraCluster
+		fixtureManipulator func(*testing.T, *framework.StateFixture)
```
munnerz: FYI, I introduced something similar in cert-manager: https://github.com/jetstack/cert-manager/pull/309/files#diff-db7d66ec1a20f75e1ea91f197b370598R27

Mine introduces a `PreFn` and ` CheckFn` which is passed the fixture structure so it can be manipulated/prepared 😄

pkg/controllers/cassandra/role/control_test.go
```diff
@@ -1,40 +1,93 @@
...
-	)
-	t.Run(
-		"sync fails",
-		func(t *testing.T) {
-			f := casstesting.NewFixture(t)
-			f.RoleControl = &casstesting.FakeControl{
-				SyncError: fmt.Errorf("simulated sync error"),
-			}
-			f.RunExpectError()
+		"no error if not yet listed": {
```
munnerz: What test case is this checking? Where the resource exists in the client, but hasn't been seen by the lister yet?

Seems to me like we *would* expect to get an error in that case, as we'd attempt to create the resource (and then it fails as it already exists). Then, the next sync will occur after a period of time long enough for the lister to be up to date.

pkg/controllers/cassandra/rolebinding/control.go
```diff
@@ -43,8 +43,15 @@ func (c *control) Sync(cluster *v1alpha1.CassandraCluster) error {
...
-		_, err = client.Create(newRoleBinding)
+	if err == nil {
+		return nil
+	}
+	if !k8sErrors.IsNotFound(err) {
+		return err
+	}
+	_, err = client.Create(newRoleBinding)
+	if k8sErrors.IsAlreadyExists(err) {
+		return nil
```
munnerz: Same comments here as in the role package 😄

pkg/controllers/cassandra/rolebinding/control_test.go
```diff
@@ -1,40 +1,93 @@
...
-			f.RunExpectError()
+		"no error if not yet listed": {
+			kubeObjects: []runtime.Object{},
+			cluster:     cluster1,
+			fixtureManipulator: func(t *testing.T, fixture *framework.StateFixture) {
+				_, err := fixture.KubeClient().RbacV1beta1().RoleBindings(rb1.Namespace).Create(rb1)
+				if err != nil {
+					t.Fatal(err)
+				}
+			},
```
munnerz: same again

pkg/controllers/executor.go
```diff
@@ -19,11 +20,14 @@ type State struct {
...
-	PodLister         corelisters.PodLister
-	ServiceLister     corelisters.ServiceLister
+	StatefulSetLister    appslisters.StatefulSetLister
+	ConfigMapLister      corelisters.ConfigMapLister
+	PilotLister          listers.PilotLister
+	PodLister            corelisters.PodLister
+	ServiceLister        corelisters.ServiceLister
+	ServiceAccountLister corelisters.ServiceAccountLister
+	RoleBindingLister    rbaclisters.RoleBindingLister
+	RoleLister           rbaclisters.RoleLister
```
munnerz: thought (not for this PR): this list is going to get long 🙃

PR #292, opened 2018-03-20
Use new unit test fixture for pilot control tests

wallrj:
* Removes the Pilot Update code. Instead we (for now) only create missing pilots.
* Later the pilot creation and deletion will be done by the ScaleOut and ScaleIn actions.
* Also makes the pilot control gracefully handle the situation where the pilot list doesn't yet have latest pilots and pilot creation therefore results in an AlreadyExists error.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cassandra/pilot/pilot.go
+++ b/pkg/controllers/cassandra/pilot/pilot.go
@@ -1,10 +1,6 @@
 package pilot
 
 import (
-	"fmt"
-	"hash/fnv"
-	"reflect"
-
 	"k8s.io/api/core/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -18,7 +14,6 @@ import (
 	navlisters "github.com/jetstack/navigator/pkg/client/listers/navigator/v1alpha1"
 	"github.com/jetstack/navigator/pkg/controllers"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra/util"
-	hashutil "github.com/jetstack/navigator/pkg/util/hash"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -79,28 +74,22 @@ func (c *pilotControl) clusterPods(cluster *v1alpha1.CassandraCluster) ([]*v1.Po
...
-	}
-	existingPilot = existingPilot.DeepCopy()
-	existingPilot.Status = v1alpha1.PilotStatus{}
-	desiredPilot = existingPilot.DeepCopy()
-	desiredPilot = updatePilotForCluster(cluster, pod, desiredPilot)
-	if !reflect.DeepEqual(desiredPilot, existingPilot) {
-		_, err = client.Update(desiredPilot)
+	_, err = client.Create(desiredPilot)
+	if k8sErrors.IsAlreadyExists(err) {
+		return nil
```
munnerz: I worry returning nil here just because a pilot with that same name already exists could cause problems. Similar to my comments in #291, I think we should return an error here and let the next invocation of the loop take the correct action (once the lister is consistent)

pkg/controllers/cassandra/pilot/pilot_test.go
```diff
@@ -32,114 +35,120 @@ func nonClusterPod(cluster *v1alpha1.CassandraCluster, name string) *v1.Pod {
...
-			f.AddObjectN(unsyncedPilot)
-			f.Run()
-			f.AssertPilotsLength(1)
-			updatedPilot := f.Pilots().Items[0]
-			updatedLabels := updatedPilot.GetLabels()
-			if updatedLabels["foo"] != "bar" {
-				t.Log(updatedLabels)
-				t.Error("custom labels were altered")
-			}
+		"no error if pilot not yet listed": {
```
munnerz: This *should* be an error IMO (we have retry logic to handle cache inconsistencies)

pkg/controllers/cassandra/pilot/pilot_test.go
```diff
@@ -32,114 +35,120 @@ func nonClusterPod(cluster *v1alpha1.CassandraCluster, name string) *v1.Pod {
...
-			f.AddObjectK(pod)
-			f.AddObjectN(unsyncedPilot)
-			f.Run()
-			f.AssertPilotsLength(1)
-			updatedPilot := f.Pilots().Items[0]
-			updatedLabels := updatedPilot.GetLabels()
-			if updatedLabels["foo"] != "bar" {
-				t.Log(updatedLabels)
-				t.Error("custom labels were altered")
-			}
```
munnerz: Is this no longer necessary?

wallrj: I've removed the update code so no longer necessary.

PR #301, opened 2018-03-27
Allow pausing clusters

kragniz:
**What this PR does / why we need it**: this allows users to 'pause' a cluster, making navigator skip syncing. This allows for manual intervention in case something goes wrong.

**Release note**:

```release-note
Allow pausing clusters
```


pkg/apis/navigator/types.go
```diff
@@ -32,6 +32,7 @@ type CassandraClusterSpec struct {
 	NodePools []CassandraClusterNodePool
 	Version   version.Version
 	Image     *ImageSpec
+	Paused    *bool
```
munnerz: Does this need to be a pointer? (i.e. is there a 3rd 'unset' state that we need to support?).

IMO `paused: false` and `paused: true` are expressive enough?

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -39,6 +39,9 @@ type CassandraClusterSpec struct {
 
 	// The version of the database to be used for nodes in the cluster.
 	Version version.Version `json:"version"`
+
+	// If set to true, no actions will take place on this cluster.
+	Paused *bool `json:"paused,omitempty"`
```
munnerz: Can this be a part of the NavigatorClusterConfig structure?

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -73,6 +73,12 @@ func NewControl(
 }
 
 func (e *defaultCassandraClusterControl) Sync(c *v1alpha1.CassandraCluster) error {
+	if c.Spec.Paused != nil && *c.Spec.Paused == true {
+		glog.V(4).Infof("defaultCassandraClusterControl.Sync skipped, since cluster is paused")
+		e.recorder.Eventf(c, apiv1.EventTypeNormal, "spec.paused", "Cluster paused, not syncing")
```
munnerz: The event name here (`spec.paused`) should be pulled into a const, and formatted similar to the other events.

munnerz: (we also need to move them into `pkg/apis/navigator/v1alpha1` at some point)

pkg/apis/navigator/types.go
kragniz: I did this so that omitempty hides the field when it's not set, but I don't think that's actually necessary. I'll change it to `bool`

pkg/controllers/elasticsearch/cluster_control.go
```diff
@@ -29,13 +29,16 @@ import (
 const (
 	errorSync = "ErrSync"
 
+	pauseField = "spec.paused"
```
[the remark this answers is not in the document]
kragniz: I was trying to be consistent with event messages like this one:

    cass-example2-ringnodes-0.151fd039c8fa1dd1   Pod       spec.initContainers{install-pilot}   Normal    Started   kubelet, minikube   Started container

But I'll change it if `Paused` is the preferred way

munnerz:
Interestingly, the deployment controller does not pause scale actions when `spec.paused` is set. It also does not log an event about it, but instead sets a `DeploymentProgressing` condition (in this function: https://github.com/kubernetes/kubernetes/blob/1102fd0dcbc4a408045e8d1bc42f056909e72322/pkg/controller/deployment/sync.go#L75).

Given the nature of a scale event in an ES or C* cluster, I think we *should* also pause scale events in the case of our `pause` feature, but maybe we want to borrow the behaviour around setting a condition/logging an event (i.e. don't log an event, instead set a condition on the resource).

We've not really utilised conditions up until now, but they do provide a nice and clear way to provide insight into the state of a resource (and save hammering the events API with duplicate events). An event probably shouldn't be used to report the state of objects, but instead used to report information on an action (perhaps not necessarily just 'Actions' as in our definition of an action) that has been taken (whether it has succeeded, or conversely if it errored then reporting some information here too). There may be some information that makes sense to duplicate as both a condition *and* an event, but I'm not sure. I guess it depends on the nature of the action.

wallrj:
Some interesting thoughts here on `PauseCondition`: https://github.com/kubernetes/kubernetes/pull/58465#issuecomment-359593540

pkg/apis/navigator/types.go
```diff
@@ -151,12 +153,51 @@ type NavigatorClusterConfig struct {
...
+	// replicas required are up and running for at least minReadySeconds.
+	ClusterAvailable ClusterConditionType = "Available"
+	// Progressing means the cluster is progressing. Progress for a deployment is
+	// considered when a new replica set is created or adopted, and when new pods scale
+	// up or old pods scale down. Progress is not estimated for paused deployments or
+	// when progressDeadlineSeconds is not specified.
+	ClusterProgressing ClusterConditionType = "Progressing"
+	// ReplicaFailure is added in a cluster when one of its pods fails to be created
+	// or deleted.
+	ClusterReplicaFailure ClusterConditionType = "ReplicaFailure"
```
munnerz: I can't see where we actually use/set these conditions?

Also, update comments (assuming they've come from Deployment defs) and remove conditions we don't currently implement.

kragniz: Actually implemented now! (I should have added wip back to the title)

docs/paused-clusters.rst
```diff
@@ -0,0 +1,5 @@
+Pausing Clusters
+----------------
+
+A cluster can be paused by setting ``spec.paused`` to ``true``.
+When this is set, navigator will not perform any actions on the cluster, allowing for manual intervention.
```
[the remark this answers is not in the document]
wallrj: I think it'd be worth adding a paragraph about the conditions you can expect to find when the cluster is paused and when it has resumed.

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -79,9 +78,52 @@ func NewControl(
 	}
 }
 
+// checkPausedConditions checks if the given cluster is paused or not and adds an appropriate condition.
+func (e *defaultCassandraClusterControl) checkPausedConditions(c *v1alpha1.CassandraCluster) error {
+	cond := c.Status.GetStatusCondition(v1alpha1.ClusterConditionProgressing)
+	pausedCondExists := cond != nil && cond.Reason == v1alpha1.PausedClusterReason
```
wallrj: Should we also look for `cond.Status == True` here?

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -79,9 +78,52 @@ func NewControl(
...
+
+	needsUpdate := false
+	if c.Spec.Paused && !pausedCondExists {
+		c.Status.UpdateStatusCondition(
+			v1alpha1.ClusterConditionProgressing,
+			v1alpha1.ConditionUnknown,
```
wallrj: Shouldn't this be `ConditionFalse` i.e. The cluster is not `progressing`

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -79,9 +78,52 @@ func NewControl(
...
+			v1alpha1.PausedClusterReason,
+			"Cluster is paused",
+		)
+		needsUpdate = true
+	} else if !c.Spec.Paused && pausedCondExists {
+		c.Status.UpdateStatusCondition(
+			v1alpha1.ClusterConditionProgressing,
+			v1alpha1.ConditionUnknown,
```
wallrj: And should this be `ConditionTrue` ?

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -79,9 +78,52 @@ func NewControl(
...
+		)
+		needsUpdate = true
+	}
+
+	if !needsUpdate {
+		return nil
+	}
+
+	var err error
+	c, err = e.state.NavigatorClientset.NavigatorV1alpha1().CassandraClusters(c.Namespace).UpdateStatus(c)
```
wallrj: nit `c` is unused.

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -79,9 +78,52 @@ func NewControl(
...
+	return err
+}
+
 func (e *defaultCassandraClusterControl) Sync(c *v1alpha1.CassandraCluster) error {
+	c = c.DeepCopy()
+	var err error
+
+	e.checkPausedConditions(c)
```
wallrj: Check and  return the error here so that the sync and checkPausedConditions will be retried in case of e.g. conflict errors.

pkg/controllers/elasticsearch/cluster_control.go
```diff
@@ -100,10 +100,50 @@ func NewController(
...
+	var err error
+	c, err = e.navigatorClient.NavigatorV1alpha1().ElasticsearchClusters(c.Namespace).UpdateStatus(c)
+	return err
+}
+
 func (e *defaultElasticsearchClusterControl) Sync(c *v1alpha1.ElasticsearchCluster) (v1alpha1.ElasticsearchClusterStatus, error) {
 	c = c.DeepCopy()
 	var err error
 
+	e.checkPausedConditions(c)
```
wallrj: And check and return error here too.

pkg/controllers/elasticsearch/cluster_control.go
```diff
@@ -100,10 +100,50 @@ func NewController(
...
+
+	if !needsUpdate {
+		return nil
+	}
+
+	var err error
+	c, err = e.navigatorClient.NavigatorV1alpha1().ElasticsearchClusters(c.Namespace).UpdateStatus(c)
+	return err
+}
+
```
wallrj: Maybe if this method instead mutated the Cluster Status in place and if there was an interface for cluster condition operations, it could be made into a generic function shared between cassandra and elasticsearch.
And the status is updated lower by `Controller.sync` lower in the stack, so we could avoid a separate API call.

Happy if you prefer to leave this for a  followup branch.

munnerz:
3/3 failed tests, even if they are on a flake, isn't a healthy state to be in. FWIW, our test infrastructure is quite over-provisioned at the moment so I doubt it's down to overload.

Perhaps we've got some other component here that's flakey?

pkg/controllers/cassandra/cluster_control.go
```diff
@@ -79,9 +78,45 @@ func NewControl(
...
+			"Cluster is paused",
+		)
+	} else if !c.Spec.Paused && pausedCondExists {
+		c.Status.UpdateStatusCondition(
+			v1alpha1.ClusterConditionProgressing,
+			v1alpha1.ConditionTrue,
+			v1alpha1.ResumedClusterReason,
+			"Cluster is resumed",
+		)
+	}
```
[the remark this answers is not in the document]
kragniz: I think this should be done by other components that progress the cluster. 

For example, the deployment controller sets conditions such as:

```
  - lastTransitionTime: 2018-05-08T14:14:46Z
    lastUpdateTime: 2018-05-08T14:14:56Z
    message: ReplicaSet "navigator-navigator-apiserver-86c554b4f" has successfully
      progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
```

maybe we should add similar conditions for each action performed on a cluster?

Issue #304, opened 2018-03-27
ParseNodePoolResourceName fails when node pool name contains a `-`


munnerz:
I've just tried deploying a Cassandra cluster and with two nodepools: `ringnodes-a` and `ringnodes-b`. It is complaining:

```
statefulset name "cass-demo-ringnodes-a" did not contain cluster name "demo"
```

We shouldn't be relying on parsing the name of the statefulset in order to determine anything. We should use labels/annotations instead. This locks us to a particular naming format forever, and if ever we look at 'rolling' statefulsets (like how the Deployment controller does with ReplicaSets), this will become a bit of a nightmare to manage 😄 

/kind bug
/area cassandra

PR #309, opened 2018-04-03
Identify statefulsets by labels rather than error prone name parsing

wallrj:
* Ensure that nodepool name label is added to the statefulsets created for each nodepool
* Use that label when creating nodepool status for each statefulset (rather than attempting to parse the ss name)
* Use more realistic nodepool names (including dashes) in the docs and in the tests.

Fixes: #304 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cassandra/nodepool/nodepool.go
+++ b/pkg/controllers/cassandra/nodepool/nodepool.go
@@ -5,8 +5,6 @@ import (
 	appslisters "k8s.io/client-go/listers/apps/v1beta1"
 	"k8s.io/client-go/tools/record"
 
-	"github.com/golang/glog"
-
 	v1alpha1 "github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
 	"github.com/jetstack/navigator/pkg/controllers/cassandra/util"
 )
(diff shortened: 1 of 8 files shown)


wallrj:
In the initial commit I used `-` in the nodepool names, recreating the original bug report. 
So these tests should all fail:
 * https://jetstack-build-infra.appspot.com/build/jetstack-logs/pr-logs/pull/jetstack_navigator/309/navigator-e2e-v1-7/1093/
 * https://jetstack-build-infra.appspot.com/build/jetstack-logs/pr-logs/pull/jetstack_navigator/309/navigator-e2e-v1-8/1107/
 * https://jetstack-build-infra.appspot.com/build/jetstack-logs/pr-logs/pull/jetstack_navigator/309/navigator-e2e-v1-9/646/

pkg/controllers/cassandra/nodepool/nodepool.go
```diff
@@ -46,14 +44,8 @@ func (e *defaultCassandraClusterNodepoolControl) updateStatus(cluster *v1alpha1.
...
-	for ssName, ss := range sets {
-		clusterName, npName, err := util.ParseNodePoolResourceName(ssName)
-		if err != nil {
-			glog.Errorf("Error parsing statefulset name: %s", err)
-		}
-		if clusterName != cluster.Name {
-			glog.Errorf("statefulset name %q did not contain cluster name %q", ssName, cluster.Name)
-		}
+	for _, ss := range sets {
+		npName := ss.Labels[util.NodePoolNameLabelKey]
```
munnerz: What if the label isn't set?

Eventually I think this sort of thing should be moved into some kind of 'meta accessor' helper... Not sure how that's best to be structured. We should dig into the upstream codebase. Doesn't need to be for this PR though.

wallrj: Yeah, I don't know where to draw the line in handling cases where an outside program or person has meddled with the labels.  If there isn't a label there'll be a `CassandraCluster.Status.NodePools[""]` NodePoolStatus entry which should give a clue to what's gone wrong.

And if we do some day change the statefulset name format we may and up with multiple sets with the same nodepool label.

I'll leave it like this for now.

Issue #265, opened 2018-02-27
Audit the API documentation


wallrj:
In https://github.com/jetstack/navigator/pull/244#discussion_r170941157 @kragniz  asked:
> should this now be marked +optional?

I'm not sure. 

Perhaps we should:
 * Check that our APIs are fully and consistently documented.
 * Maybe add CI checks for missing or incorrect API documentation
 * and we should actually generate some API documentation of course :-)

 There's a similar issue in Kubernetes with some guidelines: https://github.com/kubernetes/kubernetes/issues/14924

/kind feature

PR #311, opened 2018-04-03
Cleanup API surface

### LINKED ISSUE #194: Cleaning up v1alpha1 API in preparation for a 'stable' cut

munnerz:

Final cleanup of the API surface in prep for 0.1.

This removes the `lastCompletedPhase` field, which is currently unused (and is being set incorrectly by the Elasticsearch Pilot).

I've made many fields here omitempty, which should hopefully stop so many 'null' fields. I've also made the `Version` field for both `spec.version` and `status.{elasticsearch,cassandra}.version` a pointer so we can better represent unknown states.

**Which issue this PR fixes**: fixes #194

This PR also switches us to use pointers for node pool replica fields, and defaults them to 1 if not set. A user is still free to set it to 0 if they want an empty node pool (and if validation for that cluster type allows it)

**Release note**:
```release-note
NONE
```

/cc @wallrj


pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -32,38 +32,58 @@ type CassandraCluster struct {
 type CassandraClusterSpec struct {
 	NavigatorClusterConfig `json:",inline"`
 
-	NodePools []CassandraClusterNodePool `json:"nodePools"`
+	// List of node pools belonging to this CassandraCluster.
+	// NodePools cannot currently be removed.
+	// A cluster with 0 node pools will not function correctly.
+	NodePools []CassandraClusterNodePool `json:"nodePools,omitempty"`
```
wallrj: Not sure about this. Would an empty list be so bad? Clearer I'd say, than the attribute just not being there.
What if I need to write a jsonpath query to pick an item from that list? And the list just isn't there in the json representation.

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -32,38 +32,58 @@ type CassandraCluster struct {
...
-	Replicas int32  `json:"replicas"`
+	// Name of the node pool specified as a DNS_LABEL.
+	// Each node pool must specify a name.
+	Name string `json:"name"`
+
+	// The number of desired replicas in this node pool.
+	// This value will correspond to the number of replicas in the created
+	// StatefulSet.
+	// If not set, a default of 1 will be used.
+	Replicas *int32 `json:"replicas,omitempty"`
```
wallrj: Also not sure about this being `omitempty`.
It seems to me that it'd be better to require this to be specified.
Can / should these be uint32 to ensure that they're not negative?

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -77,10 +97,13 @@ type CassandraClusterNodePool struct {
 }
 
 type CassandraClusterStatus struct {
-	NodePools map[string]CassandraClusterNodePoolStatus `json:"nodePools"`
+	// The status of each node pool in the CassandraCluster.
+	// This will be periodically updated by Navigator.
+	NodePools map[string]CassandraClusterNodePoolStatus `json:"nodePools,omitempty"`
```
wallrj: Same comment as for `Spec.NodePools` above. I think I'd prefer this to be shown as an empty map in the serialized json than omitted.

wallrj:
See also: https://github.com/jetstack/navigator/issues/265

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -32,38 +32,58 @@ type CassandraCluster struct {
...
+	// NodePools cannot currently be removed.
+	// A cluster with 0 node pools will not function correctly.
+	NodePools []CassandraClusterNodePool `json:"nodePools,omitempty"`
 
-	// Image describes the database image to use
-	Image *ImageSpec `json:"image"`
+	// Image describes the Cassandra database image to use.
+	// This should only be set if version auto-detection is not desired.
+	// If set, the image tag used must correspond to the version specified
+	// in 'spec.version'.
```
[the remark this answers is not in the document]
munnerz: The reason for allowing a custom Image is to allow users to override the image generally, not just for non semver tags.

I think we should link off to a page in our docs that outlines how a user should configure this field, and emphasise that it's unlikely a user would want to set this/label it as an 'advanced feature'?

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -32,38 +32,58 @@ type CassandraCluster struct {
...
-	NodeSelector map[string]string `json:"nodeSelector"`
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
 	// Rack specifies the cassandra rack with which to label nodes in this
-	// nodepool. If this is not set, a default will be selected.
+	// nodepool.
+	// If this is not set, the name of the node pool will be used.
 	// +optional
-	Rack string `json:"rack"`
+	Rack string `json:"rack,omitempty"`
```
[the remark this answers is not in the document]
munnerz: Correct - I had figured there'd be no harm in putting omitempty onto this field was all.

This is a greater issue when validation is performed client side afaict (because when it is, if the value is set to "" then it won't actually be sent in the JSON payload). Given kubectl doesn't actually know about these types however, I don't think it makes a difference in our case. Happy to remove it.

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -32,38 +32,58 @@ type CassandraCluster struct {
...
 
 	// Datacenter specifies the cassandra datacenter with which to label nodes
 	// in this nodepool. If this is not set, a default will be selected.
+	// If this is not set, the default of 'navigator-default-datacenter' will
+	// be used.
 	// +optional
-	Datacenter string `json:"datacenter"`
+	Datacenter string `json:"datacenter,omitempty"`
```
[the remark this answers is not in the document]
munnerz: Same as rack attribute

pkg/apis/navigator/v1alpha1/types.go
munnerz: Lists and maps render to `null` if not set. This can break some clients, and looks ugly in `kubectl `output (and also is weird from a JSON point of view).

We have to include `omitempty` unless we have some form of validation that ensures that there's always at least one thing in the array basically. We don't currently have that check as part of our validation. Perhaps we should, but I didn't want to make that decision as part of this PR.

pkg/apis/navigator/v1alpha1/types.go
munnerz: We have omitempty here again because otherwise this will be serialised to null. Defaults will handle setting this to an appropriate value if not specified.

Having this as a pointer allows us to recognise `nil` vs `0` (so we can support 0 sized node pools). I chose to do this because every other Kubernetes workloads controller treats 'nil' to mean 1 (i.e., not specified and thus defaulting to 1).

This also helps us work with autoscalers in future (as the best practice when using an autoscaler is to not specify a replicas field in VCS, and instead allow the autoscaler to set it). If we made this a non-ptr/required, whenever a `kubectl apply` is performed, the `replicas` field would be reset to 0.

pkg/apis/navigator/v1alpha1/types.go
munnerz: That's not possible given how maps are serialised in Go. This field is also optional, as it may not always be set, hence it needs omitempty (and should also have the +optional tag)

Issue #320, opened 2018-04-10
CassandraCluster Version validation should only allow supported version (3)


wallrj:
We should add validation to give early warning to people attempting to create clusters with unsupported versions of Cassandra.

/kind bug

PR #321, opened 2018-04-10
Add validation to only allow Cassandra major version 3

wallrj:
* Only allow versions >= 3.0.0 && < 4.0.0
* Disallow version changes in API updates (pending #277)
* Document the supported versions.

Fixes: #320
**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/apis/navigator/validation/cassandra.go
+++ b/pkg/apis/navigator/validation/cassandra.go
@@ -1,14 +1,58 @@
 package validation
 
 import (
+	"fmt"
 	"reflect"
+	"strings"
 
 	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/jetstack/navigator/pkg/apis/navigator"
+	"github.com/jetstack/navigator/pkg/cassandra/version"
 )
+
+type versionSpecOp string
+
+const (
+	versionSpecGe versionSpecOp = ">="
+	versionSpecLt               = "<"
+)
+
+type versionSpec map[versionSpecOp]*version.Version
+
+var supportedCassandraVersions = versionSpec{
+	versionSpecGe: version.New("3.0.0"),
+	versionSpecLt: version.New("4.0.0"),
+}
+
+func (vs versionSpec) Check(v *version.Version) bool {
+	for op, opVersion := range vs {
+		switch op {
+		case versionSpecGe:
+			if v.LessThan(opVersion) {
+				return false
+			}
+		case versionSpecLt:
+			if opVersion.Equal(v) {
+				return false
+			}
+			if opVersion.LessThan(v) {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+func (vs versionSpec) String() string {
+	out := []string{}
+	for op, opVersion := range vs {
+		out = append(out, fmt.Sprintf("%s %s", op, opVersion))
+	}
+	return strings.Join(out, ", ")
+}
 
 func ValidateCassandraClusterNodePool(np *navigator.CassandraClusterNodePool, fldPath *field.Path) field.ErrorList {
 	// TODO: call k8s.io/kubernetes/pkg/apis/core/validation.ValidateResourceRequirements on np.Resources
@@ -26,6 +70,16 @@ func ValidateCassandraClusterUpdate(old, new *navigator.CassandraCluster) field.
 	allErrs := ValidateCassandraCluster(new)
 
 	fldPath := field.NewPath("spec")
+
+	if !new.Spec.Version.Equal(&old.Spec.Version) {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(
+				fldPath.Child("version"),
+				"cannot change the version of an existing cluster.",
+			),
+		)
+	}
 
 	npPath := fldPath.Child("nodePools")
 	for i, newNp := range new.Spec.NodePools {
@@ -60,6 +114,20 @@ func ValidateCassandraClusterUpdate(old, new *navigator.CassandraCluster) field.
 
 func ValidateCassandraClusterSpec(spec *navigator.CassandraClusterSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := ValidateNavigatorClusterConfig(&spec.NavigatorClusterConfig, fldPath)
+
+	if !supportedCassandraVersions.Check(&spec.Version) {
+		allErrs = append(
+			allErrs,
+			field.Forbidden(
+				fldPath.Child("version"),
+				fmt.Sprintf(
+					"%s is not supported. Supported versions are: %s",
+					spec.Version, supportedCassandraVersions,
+				),
+			),
+		)
+	}
+
 	npPath := fldPath.Child("nodePools")
 	allNames := sets.String{}
 	for i, np := range spec.NodePools {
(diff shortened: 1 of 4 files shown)


pkg/apis/navigator/validation/cassandra.go
```diff
@@ -1,15 +1,59 @@
...
+	versionSpecLt: version.New("4.0.0"),
+}
+
+func (vs versionSpec) Check(v *version.Version) bool {
+	for op, opVersion := range vs {
+		switch op {
+		case versionSpecGe:
+			if v.LessThan(opVersion) {
+				return false
+			}
```
munnerz: Can we invert the implementation here?

i.e. `versionSpecGe` should run `v.Greater` and `v.Equal` and return true

pkg/apis/navigator/validation/cassandra.go
```diff
@@ -1,15 +1,59 @@
...
+		case versionSpecLt:
+			if opVersion.Equal(v) {
+				return false
+			}
+			if opVersion.LessThan(v) {
```
munnerz: We switch to testing `opVersion` instead of `v` here - can we be consistent for readability?

pkg/apis/navigator/validation/cassandra.go
```diff
@@ -1,15 +1,59 @@
...
+	}
+	return true
+}
+
+func (vs versionSpec) String() string {
+	out := []string{}
+	for op, opVersion := range vs {
+		out = append(out, fmt.Sprintf("%s %s", op, opVersion))
+	}
+	return strings.Join(out, ", ")
```
munnerz: Maps are not consistently ordered, so this version string will be unstable. Can we make it stable?

pkg/apis/navigator/validation/cassandra.go
```diff
@@ -1,15 +1,59 @@
...
+)
+
+type versionSpec map[versionSpecOp]*version.Version
+
+var supportedCassandraVersions = versionSpec{
+	versionSpecGe: version.New("3.0.0"),
+	versionSpecLt: version.New("4.0.0"),
+}
+
+func (vs versionSpec) Check(v *version.Version) bool {
```
munnerz: Maybe a quick unit test for this function?

pkg/apis/navigator/validation/cassandra.go
wallrj: I deleted this code and instead use github.com/hashicorp/go-version constraints.

pkg/apis/navigator/validation/cassandra.go
wallrj: I deleted this code and instead use github.com/hashicorp/go-version constraints.

pkg/apis/navigator/validation/cassandra.go
wallrj: I deleted this code and instead use github.com/hashicorp/go-version constraints.

pkg/apis/navigator/validation/cassandra.go
wallrj: Switched to github.com/hashicorp/go-version constraints which already has unit tests.

wallrj:
Hey @munnerz.

Please take another look.

* I replaced my home grown constraints checking code with github.com/hashicorp/go-version constraints.
* Which seems to be more popular and better tested than the alternative https://github.com/Masterminds/semver
* Also used Hashicorp version instead of coreos semver in the Cassandra Version type.
* And added generated DeepCopy functions to that, as discussed.

I can follow this up with PRs to 
 * use the same custom Version type for Elastic search.
 * have the controller perform an additional check for supported versions to handle the case where the API server may have been upgraded but the controller has not.

pkg/cassandra/version/version.go
```diff
@@ -37,10 +37,28 @@ func New(s string) *Version {
 	return v
 }
 
+func (v *Version) set(s string) error {
+	sv, err := semver.NewVersion(s)
+	if err != nil {
+		return err
+	}
+	v.versionString = s
```
munnerz: Do we need to keep the versionString in memory? Surely `semver.String` will be more up to date?

pkg/cassandra/version/version.go
```diff
@@ -2,11 +2,9 @@ package version
 
 import (
 	"encoding/json"
-	"fmt"
 	"strconv"
-	"strings"
 
-	"github.com/coreos/go-semver/semver"
+	semver "github.com/hashicorp/go-version"
 )
```
munnerz: Can we move this file out of `pkg/cassandra` and into something like `pkg/api/util`? This is an API type, but it isn't by default part of an API group (hence it isn't in apis).

It'd be good to denote it as such in the file structure, to make it clear this isn't specific to just cassandra (and also denotes that this type may be exposed in an api)

pkg/cassandra/version/version_test.go
```diff
@@ -91,3 +93,18 @@ func TestUnmarshalJSON(t *testing.T) {
 		)
 	}
 }
+
+func TestDeepCopy(t *testing.T) {
+	t.Run(
+		"zero value",
+		func(t *testing.T) {
+			t.Log(version.Version{}.DeepCopy())
```
munnerz: This doesn't test anything, or ever fail afaict

pkg/cassandra/version/version.go
```diff
@@ -49,62 +67,20 @@ func (v *Version) UnmarshalJSON(data []byte) error {
...
-func (v Version) Semver() string {
-	return v.semver.String()
+// DeepCopy returns a deep-copy of the Version value.
+// If the underlying semver is a nil pointer, assume that the zero value is being copied,
+// and return that.
+func (v Version) DeepCopy() Version {
+	if v.semver == nil {
+		return Version{}
+	}
+	return *New(v.String())
```
munnerz: :+1:

Gopkg.toml
```diff
@@ -52,3 +52,7 @@ required = [
 [[override]]
   name = "k8s.io/gengo"
   revision = "b58fc7edb82e0c6ffc9b8aef61813c7261b785d4"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/hashicorp/go-version"
```
munnerz: Don't think this needs to be added as a constraint given we are constraining it to master anyway. A simple `dep ensure` without adding this block should yield the same result, sans this change

pkg/cassandra/version/version.go
wallrj: We need it because we want to remember the format of the originally supplied version string.
So if the Cassandra node reports version `3.11` via nodetool and a JMX query, we want to report that exact version in the `Pilot.Status.Cassandra.Version` rather than the semver `3.11.0`.

We also don't want a user to submit `CassandraCluster.Spec.Version: 3.11` and then have that change to `3.11.0` when they do `kubectl get CassandraCluster -o yaml`.

However, when we infer the Docker image tag, we use the semver, because we don't want docker to simply pull the latest `3.11` image from Dockerhub.

pkg/cassandra/version/version_test.go
wallrj: It tests that a Zero value can be DeepCopied.
It originally caused a panic before I added a check for nil pointer in the implementation.
I'll leave it for now.

pkg/cassandra/version/version.go
wallrj: Done.

Gopkg.toml
wallrj: Done.

Issue #322, opened 2018-04-10
Move Cassandra pilot creation to ScaleOut 


wallrj:
I think it make sense to create pilots during scaleout rather than as a separate control.

And then in #285 we can remove the pilot during scale down.

> /kind feature

wallrj:
We're trying a generic pilot controller instead: https://github.com/jetstack/navigator/pull/353

PR #328, opened 2018-04-11
Delete left over pilots

wallrj:
Contrary to the original idea of creating pilots during scale-out and deleting them during scale-in,
I've instead extended the existing pilot sync method to delete pilots with an index higher than the CurrentReplicas value of the corresponding StatefulSet.
And I modified the pilot creation logic to select pods based on labels rather than based on cluster > statefulset ownership.

Fixes: #322

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/apis/navigator/v1alpha1/types.go
+++ b/pkg/apis/navigator/v1alpha1/types.go
@@ -15,8 +15,9 @@ const (
 	ElasticsearchNodePoolVersionAnnotation = "navigator.jetstack.io/elasticsearch-version"
 	ElasticsearchRoleLabelPrefix           = "navigator.jetstack.io/elasticsearch-role-"
 
-	CassandraClusterNameLabel  = "navigator.jetstack.io/cassandra-cluster-name"
-	CassandraNodePoolNameLabel = "navigator.jetstack.io/cassandra-node-pool-name"
+	CassandraClusterNameLabel   = "navigator.jetstack.io/cassandra-cluster-name"
+	CassandraNodePoolNameLabel  = "navigator.jetstack.io/cassandra-node-pool-name"
+	CassandraNodePoolIndexLabel = "navigator.jetstack.io/cassandra-node-pool-index"
 )
 
 // +genclient
--- a/pkg/controllers/cassandra/pilot/pilot.go
+++ b/pkg/controllers/cassandra/pilot/pilot.go
@@ -1,6 +1,10 @@
 package pilot
 
 import (
+	"strconv"
+	"strings"
+
+	"k8s.io/api/apps/v1beta1"
 	"k8s.io/api/core/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -8,6 +12,7 @@ import (
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/record"
 
+	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
 	"github.com/jetstack/navigator/pkg/apis/navigator/v1alpha1"
@@ -89,7 +94,59 @@ func (c *pilotControl) Sync(cluster *v1alpha1.CassandraCluster) error {
 			return errors.Wrap(err, "unable to ensure pilot")
 		}
 	}
+
+	sets, err := c.statefulSets.StatefulSets(cluster.Namespace).List(selector)
+	if err != nil {
+		return errors.Wrap(err, "unable to list statefulsets")
+	}
+
+	setsByNodePoolName := map[string]*v1beta1.StatefulSet{}
+	for _, set := range sets {
+		setNodePoolName := set.Labels[v1alpha1.CassandraNodePoolNameLabel]
+		setsByNodePoolName[setNodePoolName] = set
+	}
+
+	pilots, err := c.pilots.Pilots(cluster.Namespace).List(selector)
+	if err != nil {
+		return errors.Wrap(err, "unable to list pilots")
+	}
+
+	for _, pilot := range pilots {
+		pilotNodePoolName := pilot.Labels[v1alpha1.CassandraNodePoolNameLabel]
+		setForPilot := setsByNodePoolName[pilotNodePoolName]
+		pilotIndexLabel := pilot.Labels[v1alpha1.CassandraNodePoolIndexLabel]
+		pilotIndex, err := strconv.Atoi(pilotIndexLabel)
+		if err != nil {
+			glog.Errorf(
+				"Unable to parse pilot %s/%s index: %q",
+				pilot.Namespace, pilot.Name, pilotIndexLabel,
+			)
+		}
+		if int32(pilotIndex) > setForPilot.Status.CurrentReplicas {
+			err := c.naviClient.NavigatorV1alpha1().
+				Pilots(cluster.Namespace).Delete(pilot.Name, &metav1.DeleteOptions{})
+			if err != nil {
+				if !k8sErrors.IsNotFound(err) {
+					return errors.Wrapf(
+						err, "unable to delete pilot %s/%s", pilot.Namespace, pilot.Name,
+					)
+				}
+			}
+		}
+	}
 	return nil
+}
+
+func parsePodIndex(podName string) (int, bool) {
+	parts := strings.Split(podName, "-")
+	if len(parts) < 2 {
+		return -1, false
+	}
+	index, err := strconv.Atoi(parts[len(parts)])
+	if err != nil {
+		return -1, false
+	}
+	return index, true
 }
 
 func PilotForCluster(cluster *v1alpha1.CassandraCluster, pod *v1.Pod) *v1alpha1.Pilot {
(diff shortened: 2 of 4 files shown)


munnerz:
I'm not sure we should be using currentReplicas here:

```
FIELD: currentReplicas <integer>

DESCRIPTION:
     currentReplicas is the number of Pods created by the StatefulSet controller
     from the StatefulSet version indicated by currentRevision.
```

During an update of the statefulset, this number will represent the total number of *out of date* replicas (i.e. it will gradually reduce from N to 0 during the upgrade, and then once the upgrade is complete will reset back to N).

FWIW, here are the various StatefulSet.status fields:

```
FIELDS:
   collisionCount	<integer>
     collisionCount is the count of hash collisions for the StatefulSet. The
     StatefulSet controller uses this field as a collision avoidance mechanism
     when it needs to create the name for the newest ControllerRevision.

   conditions	<[]Object>
     Represents the latest available observations of a statefulset's current
     state.

   currentReplicas	<integer>
     currentReplicas is the number of Pods created by the StatefulSet controller
     from the StatefulSet version indicated by currentRevision.

   currentRevision	<string>
     currentRevision, if not empty, indicates the version of the StatefulSet
     used to generate Pods in the sequence [0,currentReplicas).

   observedGeneration	<integer>
     observedGeneration is the most recent generation observed for this
     StatefulSet. It corresponds to the StatefulSet's generation, which is
     updated on mutation by the API Server.

   readyReplicas	<integer>
     readyReplicas is the number of Pods created by the StatefulSet controller
     that have a Ready Condition.

   replicas	<integer> -required-
     replicas is the number of Pods created by the StatefulSet controller.

   updateRevision	<string>
     updateRevision, if not empty, indicates the version of the StatefulSet used
     to generate Pods in the sequence [replicas-updatedReplicas,replicas)

   updatedReplicas	<integer>
     updatedReplicas is the number of Pods created by the StatefulSet controller
     from the StatefulSet version indicated by updateRevision.
```

I'd guess either `replicas`, or `currentReplicas+updatedReplicas` is what we need to use?

We could alternatively always base it on the `statefulset.spec.replicas` field - as once the StatefulSet has been scaled down, the cluster itself should already be in a state whereby it can scale down (i.e. all data evacuated from the node). Perhaps I've missed an edge case here though?

pkg/controllers/cassandra/pilot/pilot_test.go
```diff
@@ -1,45 +1,56 @@
...
+	"fmt"
 	"testing"
 
+	"k8s.io/api/apps/v1beta1"
 	"k8s.io/api/core/v1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
+	"github.com/kr/pretty"
```
munnerz: Can you fix up grouping of imports here?

```
import (
//stdlib

//external pkgs

//navigator pkgs
)
```

pkg/controllers/cassandra/pilot/pilot_test.go
```diff
@@ -1,45 +1,56 @@
...
+	cluster1np1pilot0foreign := cluster1np1pilot0.DeepCopy()
+	cluster1np1pilot0foreign.SetOwnerReferences([]metav1.OwnerReference{})
 
 	cluster2 := casstesting.ClusterForTest()
 	cluster2.SetName("cluster2")
 	cluster2.SetUID("uid2")
-	cluster2pod1 := clusterPod(cluster2, "c2p1")
+	cluster2np1 := &cluster2.Spec.NodePools[0]
+	cluster2np1set := nodepool.StatefulSetForCluster(cluster2, cluster2np1)
+	cluster2np1pod0 := clusterPod(cluster2, cluster2np1set, 0)
```
munnerz: It'd be great if we can avoid such a huge stack of objects declared here. It's really difficult when looking at a particular test case to see what's being tested (i.e. the inputs and outputs).

Not sure if there's an easy way to do this, but right now I am definitely struggling to understand all this.

I managed to get around this in cert-manager by accepting a pointer to the fixture itself as part of a 'PreFn': https://github.com/jetstack/cert-manager/blob/master/pkg/issuer/acme/http/ingress_test.go#L21-L54

If not, comments throughout all of this may help make this more readable 😄

pkg/controllers/cassandra/util/util.go
```diff
@@ -66,6 +68,35 @@ func SelectorForCluster(c *v1alpha1.CassandraCluster) (labels.Selector, error) {
...
+func SelectorForClusterNodePools(c *v1alpha1.CassandraCluster, nodePoolNames ...string) (labels.Selector, error) {
+	selector, err := SelectorForCluster(c)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to create base selector")
+	}
+	nodePoolNameExistsRequirement, err := labels.NewRequirement(
+		v1alpha1.CassandraNodePoolNameLabel,
+		selection.Exists,
+		nil,
+	)
```
munnerz: Flatten to a single line

munnerz: Also why is this being set to nil? Doesn't make sense to me? 🤔

pkg/controllers/cassandra/util/util.go
```diff
@@ -66,6 +68,35 @@ func SelectorForCluster(c *v1alpha1.CassandraCluster) (labels.Selector, error) {
 	return labels.NewSelector().Add(*clusterNameReq), nil
 }
 
+func SelectorForClusterNodePools(c *v1alpha1.CassandraCluster, nodePoolNames ...string) (labels.Selector, error) {
```
munnerz: Brief comment would be great

pkg/controllers/cassandra/util/util.go
wallrj: `Exists` selections can not have a value.

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -51,76 +56,139 @@ func NewControl(
...
+		if int32(pilotIndex+1) <= setForPilot.Status.CurrentReplicas {
+			continue
+		}
+
+		_, err = c.pods.Pods(cluster.Namespace).Get(pilot.Name)
+		if err == nil {
+			glog.V(4).Infof(
+				"Not deleting pilot %s/%s because its pod still exists.",
+				pilot.Namespace, pilot.Name,
+			)
```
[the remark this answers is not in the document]
wallrj: Yeah, but my editor doesn't wrap these lines (nor does Github in its web based diffs).
I find it much easier to read the code this way.

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -51,76 +56,139 @@ func NewControl(
...
+			continue
+		}
+		if !k8sErrors.IsNotFound(err) {
+			return errors.Wrap(err, "unable to get pod for pilot")
+		}
+		err = c.naviClient.NavigatorV1alpha1().
+			Pilots(cluster.Namespace).Delete(pilot.Name, &metav1.DeleteOptions{})
+		if err == nil {
+			glog.V(4).Infof("Deleted pilot %s/%s.", pilot.Namespace, pilot.Name)
+		} else {
```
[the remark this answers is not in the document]
wallrj: Agreed. Done.

pkg/controllers/cassandra/pilot/pilot.go
```diff
@@ -51,76 +56,139 @@ func NewControl(
...
+					err, "unable to delete pilot %s/%s", pilot.Namespace, pilot.Name,
+				)
+			}
+		}
 	}
-	// TODO: Housekeeping. Remove pilots that don't have a corresponding pod.
 	return nil
 }
 
+func parsePodIndex(podName string) (int, bool) {
```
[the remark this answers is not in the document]
wallrj: I removed this function, it's no longer needed.

pkg/controllers/cassandra/pilot/pilot_test.go
wallrj: With the revised implementation I found I could remove many of these.
I tried to make the variable names self-explanatory. 
Not sure what more I can add with comments.

pkg/apis/navigator/v1alpha1/types.go
```diff
@@ -15,8 +15,9 @@ const (
 	ElasticsearchNodePoolVersionAnnotation = "navigator.jetstack.io/elasticsearch-version"
 	ElasticsearchRoleLabelPrefix           = "navigator.jetstack.io/elasticsearch-role-"
 
-	CassandraClusterNameLabel  = "navigator.jetstack.io/cassandra-cluster-name"
-	CassandraNodePoolNameLabel = "navigator.jetstack.io/cassandra-node-pool-name"
+	CassandraClusterNameLabel   = "navigator.jetstack.io/cassandra-cluster-name"
+	CassandraNodePoolNameLabel  = "navigator.jetstack.io/cassandra-node-pool-name"
+	CassandraNodePoolIndexLabel = "navigator.jetstack.io/cassandra-node-pool-index"
```
munnerz: This appears to be set, but not actually used (at least as part of this PR).

It also could introduce some issues around generalising this code between controllers if we ever have a controller that doesn't use a StatefulSet.

Can we remove it for now in favour of discussion at a later date of storing the index vs the pod name on the Pilot?

PR #329, opened 2018-04-12
Use ordered pod management

wallrj:
Our ScaleOut action already adds pods one-at-a-time, but using the parallel management policy in the statefulset was definitely wrong.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/docs/cassandra.rst
+++ b/docs/cassandra.rst
@@ -227,4 +227,13 @@ Scale Out
 
 When you first create a cluster or when you increment the ``CassandraCluster.Spec.NodePools[i].ReplicaCount``,
 Navigator will add C* nodes, one at a time, until the desired number of nodes is reached.
+
+.. note::
+
+   Navigator adds C* Nodes in series (one-at-a-time)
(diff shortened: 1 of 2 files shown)


wallrj:
> How come we can't use a parallel strategy? Does cassandra recommend nodes be brought online one at a time?

There are mixed messages about this. The official docs don't really mention it:
 * https://docs.datastax.com/en/cassandra/3.0/cassandra/operations/opsAddNodeToCluster.html?hl=add%2Cnode

But elsewhere it is implied that adding nodes, one-at-a-time is important when using the `auto_bootstrap=true` configuration option. E.g. this question and the answer from https://stackoverflow.com/users/1054558/aaron (top rated for Cassandra https://stackoverflow.com/tags/cassandra/topusers):
https://stackoverflow.com/questions/37283424/best-way-to-add-multiple-nodes-to-existing-cassandra-cluster

> In some cases, could waiting until each node is Ready before proceeding cause a cluster to not start due to a lack of quorum, causing the first pods to never become Ready?

I don't think so. Adding nodes in parallel might cause problems for clients reads and writes where they require quorum consistency and  data has not yet replicated to the new nodes. 

I'll add some notes and links to the docs.

wallrj:
Added a note to the docs.

PR #332, opened 2018-04-14
Use the pure Java kubernetes-seedprovider

wallrj:
* from before https://github.com/kubernetes/examples/pull/201
* because that introduced a new Go / CGo shared library dependency which we don't want to have to bundle into our pilot image.

**Release note**:
```release-note
NONE
```


kragniz:
:+1: , I suggest we fork and maintain it ourselves at this point.

Issue #331, opened 2018-04-12
Error: mapped nodes must be a superset of Live and Unreachable nodes


wallrj:
This results in existing healthy nodes failing their readiness probe when new nodes are added  to the cluster.

I think this is a mistake in my understanding of these fields.
I assumed that a node must be one of Live or Unreachable,

And that all the nodes are listed in the mapped array.

But in the example below, the "live" node 10.192.3.19 does not appear in the mapped list.

```
E0412 13:45:01.694417       1 listen.go:25] Error while running Check function for probe on port 12000: mapped nodes must be a superset of Live and Unreachable nodes. Live: map[10.192.3.18:{} 10.192.3.19:{} 10.192.2.8:{} 10.192.3.15:{} 10.192.3.16:{} 10.192.3.17:{} 10.192.2.6:{} 10.192.2.7:{} 10.192.2.9:{}], Unreachable: map[], Mapped: map[10.192.3.17:{} 10.192.2.6:{} 10.192.3.18:{} 10.192.3.15:{} 10.192.3.16:{} 10.192.2.8:{} 10.192.2.7:{} 10.192.2.9:{}]
```

```
Live:
map[
10.192.2.6:{}
10.192.2.7:{}
10.192.2.8:{}
10.192.2.9:{}
10.192.3.15:{}
10.192.3.16:{}
10.192.3.17:{}
10.192.3.18:{}
10.192.3.19:{}
],

Unreachable:
map[],

Mapped:

map[
10.192.2.6:{}
10.192.2.7:{}
10.192.2.8:{}
10.192.2.9:{}
10.192.3.15:{}
10.192.3.16:{}
10.192.3.17:{}
10.192.3.18:{}
]
```

I'll double check what the real nodetool does in this case.

/kind bug

PR #333, opened 2018-04-14
Remove broken host map subset assertions from nodetool

wallrj:
Fixes: #331

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cassandra/nodetool/nodetool.go
+++ b/pkg/cassandra/nodetool/nodetool.go
@@ -23,7 +23,7 @@ const (
 	NodeStateMoving  NodeState = "Moving"
 )
 
-// NodeState represents the state of reachability of of a C* node from the
+// NodeState represents the reachability of a C* node from the
 // perspective of the node answering the query.
 type NodeStatus string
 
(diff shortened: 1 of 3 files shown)


pkg/cassandra/nodetool/nodetool.go
```diff
@@ -135,17 +124,6 @@ func (t *tool) Status() (NodeMap, error) {
...
-		return nil, fmt.Errorf(
-			"mapped nodes must be a superset of leaving, joining and moving nodes. "+
-				"Leaving: %v, "+
-				"Joining: %v, "+
-				"Moving: %v, "+
-				"Mapped: %v",
-			leavingNodes, joiningNodes, movingNodes, mappedNodes,
-		)
-	}
-
```
munnerz: This function confuses me a lot, and I'm not really sure how it works.

Can you add a comment to the function itself explaining the rough algorithm in place here, so that this change can be reviewed properly?

wallrj: I've documented the function. Please take another look.

Issue #319, opened 2018-04-10
Investigate whether Cassandra nodes need stable IP addresses


wallrj:
It's unclear how Cassandra > v3 handles node IP address changes.

There are discussions in [Sticky IPs for StatefulSet](https://github.com/kubernetes/kubernetes/issues/28969) which suggest that Cassandra does need "sticky IP addresses". But it's not clear if that applies to v2 or v3 Cassandra (or both).

On the other hand, there is [cassandra-kubernetes-hostid
](https://github.com/mstump/cassandra-kubernetes-hostid) written as part of [Improve Cassandra Example ](https://github.com/kubernetes/kubernetes/issues/34978), which suggests that you can supply the "hostid" to `-Dcassandra.replace_address` when starting the C* node.
The "hostid" seems to be a  UUID chosen by Cassandra when it first creates its data store and which is the identity of this node in the cluster, rather than the IP address.

> The best practices for Cassandra state that you should use local storage physically attached to the machine. However, if this machine fails and data is lost then you'll need to provision a replacement node. The method for doing this is to pass the IP or host UUID via the JVM flag: -Dcassandra.replace_address=e2a79390-c458-4387-8034-c14da6d38a22. Because IPs in Kubernetes are not stable the host ID is the only way to retain identity after data loss. This tool will persist or read the host ID using the annotations of the Kubernetes API.

There is a `replace_address_first_boot` option described in [Bootstrapping Apache Cassandra Nodes](http://thelastpickle.com/blog/2017/05/23/auto-bootstrapping-part1.html) which sounds like it is probably the correct option to use.

Some documentation talks about supplying an IP address to `replace_address`:
 * https://docs.datastax.com/en/cassandra/3.0/cassandra/operations/opsReplaceLiveNode.html?hl=replace_address

Other documentation talks about supplying the `listen_address` to `replace_address`:
 * https://docs.datastax.com/en/cassandra/3.0/cassandra/tools/toolsCUtility.html#toolsCUtility__cRplAdsress

And the documentation for `listen_address` talks about Ip address or hostname:
 * https://docs.datastax.com/en/cassandra/3.0/cassandra/configuration/configCassandra_yaml.html#configCassandra_yaml__listen_address

So perhaps we can simply supply the StatefulSet stable hostname to `listen_address` and that will be sufficient?

Other useful links:
 * https://mrcalonso.com/2016/01/26/cassandra-instantaneous-in-place-node-replacement/
[...]

PR #330, opened 2018-04-12
WIP: Use hostnames rather than IP addresses for cassandra nodes

wallrj:
Fixes: #319 

**Release note**:
```release-note
NONE
```

PR #334, opened 2018-04-14
Configure Cassandra nodes to lookup their IP address by their fully qualified domain name

wallrj:
* Also ensure that they only ever use the Kubernetes seed provider service
* And ensure that that service publishes the seed IP addresses as soon as they are known,
* so that seeds them selves can find themselves when they are starting up.

This is a cut down version of #330 
* without the headless service per statefulset, which I think may be unnecessary, since nodes only have to be able to lookup their own FQDN.
* and without the E2E tests, because I haven't yet figured out how to force minikube to change the IP address of a pod when it's recreated.

Fixes: #319

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cassandra/nodepool/resource.go
+++ b/pkg/controllers/cassandra/nodepool/resource.go
@@ -57,6 +57,7 @@ func StatefulSetForCluster(
 				Type: apps.RollingUpdateStatefulSetStrategyType,
 			},
 			PodManagementPolicy: apps.OrderedReadyPodManagement,
+			ServiceName:         statefulSetName,
 			Template: apiv1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: nodePoolLabels,
@@ -194,6 +195,34 @@ func StatefulSetForCluster(
 								{
 									Name:  "HEAP_NEWSIZE",
 									Value: "100M",
+								},
+								// Deliberately set to a single space to force Cassandra to do a host name lookup.
+								// See https://github.com/apache/cassandra/blob/cassandra-3.11.2/conf/cassandra.yaml#L592
+								{
+									Name:  "CASSANDRA_LISTEN_ADDRESS",
+									Value: " ",
+								},
+								{
+									Name:  "CASSANDRA_BROADCAST_ADDRESS",
+									Value: " ",
+								},
+								{
+									Name:  "CASSANDRA_RPC_ADDRESS",
+									Value: " ",
+								},
+								// Set a non-existent default seed.
+								// The Kubernetes Seed Provider will fall back to a default seed host if it can't look up seeds via the CASSANDRA_SERVICE.
+								// And if the CASSANDRA_SEEDS environment variable is not set, it defaults to localhost.
+								// Which could cause confusion if a non-seed node is temporarily unable to lookup the seed nodes from the service.
+								// We want the list of seeds to be strictly controlled by the service.
+								// See:
+								// https://github.com/docker-library/cassandra/blame/master/3.11/docker-entrypoint.sh#L31 and
+								// https://github.com/apache/cassandra/blob/cassandra-3.11.2/conf/cassandra.yaml#L416 and
+								// https://github.com/kubernetes/examples/blob/cabf8b8e4739e576837111e156763d19a64a3591/cassandra/java/src/main/java/io/k8s/cassandra/KubernetesSeedProvider.java#L69 and
+								// https://github.com/kubernetes/examples/blob/cabf8b8e4739e576837111e156763d19a64a3591/cassandra/go/main.go#L51
+								{
+									Name:  "CASSANDRA_SEEDS",
+									Value: "black-hole-dns-name",
 								},
 								{
 									Name:  "CASSANDRA_ENDPOINT_SNITCH",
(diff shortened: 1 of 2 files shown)


munnerz:
> and without the E2E tests, because I haven't yet figured out how to force minikube to change the IP address of a pod when it's recreated.

When a pod is deleted, in 99% of cases it should have a new IP address. Can we make a test that deletes the pod and verifies the IP has changed? (perhaps we delete the pod again if it hasn't changed first time?).

We definitely need an e2e test for this - ideally the test would be labelled "Resiliency" too, as the first of a suite of resiliency tests.

pkg/cassandra/nodetool/nodetool.go
```diff
@@ -135,17 +124,6 @@ func (t *tool) Status() (NodeMap, error) {
...
-		return nil, fmt.Errorf(
-			"mapped nodes must be a superset of leaving, joining and moving nodes. "+
-				"Leaving: %v, "+
-				"Joining: %v, "+
-				"Moving: %v, "+
-				"Mapped: %v",
-			leavingNodes, joiningNodes, movingNodes, mappedNodes,
-		)
-	}
-
```
munnerz: I'm confused - isn't this meant to be a part of another PR?

If yes and this is just stacked, it'd be great if we can stop stacking PRs too much as it makes it far harder to know which to review first. If we keep them separate, PRs that depend on [some other] PR will fail and/or we can add `/hold` and comment to inform a reviewer that "this PR depends on PR xyz being merged first"

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -194,6 +195,34 @@ func StatefulSetForCluster(
...
+									Value: " ",
+								},
+								{
+									Name:  "CASSANDRA_BROADCAST_ADDRESS",
+									Value: " ",
+								},
+								{
+									Name:  "CASSANDRA_RPC_ADDRESS",
+									Value: " ",
+								},
```
munnerz: Do we use the Cassandra RPC interface, and do we want to listen on the default pod IP for RPC connections? It seems to me like we'd only want to listen on 127.0.0.1?

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -194,6 +195,34 @@ func StatefulSetForCluster(
...
+								// Set a non-existent default seed.
+								// The Kubernetes Seed Provider will fall back to a default seed host if it can't look up seeds via the CASSANDRA_SERVICE.
+								// And if the CASSANDRA_SEEDS environment variable is not set, it defaults to localhost.
+								// Which could cause confusion if a non-seed node is temporarily unable to lookup the seed nodes from the service.
```
munnerz: What would happen if we were to set this to localhost instead? Would the node fail to boot, or would it cause a split brain of some sort?

pkg/controllers/cassandra/nodepool/resource.go
```diff
@@ -194,6 +195,34 @@ func StatefulSetForCluster(
...
+								// We want the list of seeds to be strictly controlled by the service.
+								// See:
+								// https://github.com/docker-library/cassandra/blame/master/3.11/docker-entrypoint.sh#L31 and
+								// https://github.com/apache/cassandra/blob/cassandra-3.11.2/conf/cassandra.yaml#L416 and
+								// https://github.com/kubernetes/examples/blob/cabf8b8e4739e576837111e156763d19a64a3591/cassandra/java/src/main/java/io/k8s/cassandra/KubernetesSeedProvider.java#L69 and
+								// https://github.com/kubernetes/examples/blob/cabf8b8e4739e576837111e156763d19a64a3591/cassandra/go/main.go#L51
+								{
+									Name:  "CASSANDRA_SEEDS",
+									Value: "black-hole-dns-name",
```
munnerz: I'm slightly concerned this could become an attack vector. What if someone creates a service named black-hole-dns-name? Could an attacker gain access to the contents of the Cassandra DB as a result?

If possible, I'd rather set this to a value that causes Cassandra to fail hard/crash (as this value *should* only be referenced in the event of a DNS resolution failure)

wallrj: I think the answer is to remove the fallback code from the Kubernetes Seed Provider.
If it can't lookup the seeds from the Kubernetes it should log an error and return an empty list,
rather than returning a default.

pkg/controllers/cassandra/nodepool/resource.go
wallrj: We don't use it, but I'm not sure whether Cassandra nodes talk to each other with this protocol.
The default is to listen on all addresses:
 * https://github.com/docker-library/cassandra/blob/master/3.11/docker-entrypoint.sh#L29

So setting it here may not be ideal, but I can improve or remove it later.

munnerz: 👍

pkg/controllers/cassandra/nodepool/resource.go
wallrj: I think it would mean that new nodes might consider themselves seeds in the event that the KubernetesSeedProvider plugin encounters an error.
So yeah, a split brain situation I think.

pkg/controllers/cassandra/nodepool/resource.go
wallrj: I'll do this in a follow up branch, where I remove the reliance on the Docker image entrypoint and instead create a Navigator default config.

Issue #346, opened 2018-05-08
Unable to use non-dockerhub Cassandra images due to dependence on the Dockerhub image entry point script


wallrj:
The Cassandra Pilot is currently hard coded to execute the Dockerhub image entry point script.

https://github.com/docker-library/cassandra/blob/master/3.11/docker-entrypoint.sh

This allows us to use environment variables to control the configuration of the Cassandra process.

But it also locks us to using that image.

Instead, we might bundle our own default Cassandra configuration file and execute `cassandra` directly.

In future we might provide a mechanism for merging the default configuration file, supplied in the image, with the configuration variables that we need to override for Navigator to manage the cluster.

/kind bug

Issue #251, opened 2018-02-15
Pattern for adjusting database configuration


kragniz:
/kind feature

Currently, there is no maintainable way to adjust the initial cassandra configuration beyond the options exposed as environment variables in the docker image we use. #245 requires using one of these extra options, which it currently does using a string replacement.

We should have a better way of making these modifications.

PR #347, opened 2018-05-08
Pilot executes `cassandra` directly, bypassing the container image entry point

wallrj:
* Cassandra Pilot now executes `cassandra` directly, bypassing the Docker image entrypoint
* Added a configuration API with which to override configuration values in yaml and properties config files.
* This replaces the experimental search / replace code for seed_provider and snitch.
* This also the use of environment variables to configure the Cassandra node. Instead, those configuration options are now supplied as Pilot command line arguments and are written to the config files in a pre-start pilot hook.
* Added an integration test so that i can see this mechanism working, without having to wait for e2e tests.

Fixes:  #346, #251 

**Release note**:
```release-note
NONE
```


pkg/pilot/genericpilot/options.go
```diff
@@ -179,5 +180,7 @@ func (o *Options) Pilot() (*GenericPilot, error) {
 func (o *Options) AddFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.PilotName, "pilot-name", "", "The name of this Pilot. If not specified, an auto-detected name will be used.")
 	flags.StringVar(&o.PilotNamespace, "pilot-namespace", "", "The namespace the corresponding Pilot resource for this Pilot exists within.")
+	flags.BoolVar(&o.LeaderElect, "leader-elect", true, ""+
```
munnerz, line 183: I'm pretty nervous exposing this option to users - it should be defined by the *Pilot* (e.g. elasticsearch-pilot etc) whether leader election needs to be performed.
Disabling this could cause fundamental breakages to how the Pilot operates, so I don't think it should be a user exposed option.

Is this because you don't use the leader election primitives in the cassandra pilot? If so, can you change this to only be configurable by the Pilot code (i.e. not a user facing option)?

internal/test/util/testfs/testfs.go
```diff
@@ -0,0 +1,53 @@
...
+	require.NoError(tfs.t, err)
+	return outPath
+}
+
+func (tfs *TestFs) TempDir(name string) string {
+	outPath := path.Join(tfs.d, name)
+	err := os.MkdirAll(outPath, os.ModePerm)
+	require.NoError(tfs.t, err)
+	return outPath
+}
```
munnerz, line 53: What's the reasoning for this package? It seems to wrap `ioutil`, but with some extras in there?
It'd be great if we can *just* use ioutil for this, to save having additional code to maintain.

https://golang.org/pkg/io/ioutil/#TempDir

internal/test/util/testfs/testfs.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type TestFs struct {
+	t *testing.T
+	d string
+}
+
+func New(t *testing.T) *TestFs {
+	d := fmt.Sprintf(".test/%s", t.Name())
```
munnerz, line 20: Is there a reason for specifying a custom base, instead of relying on `ioutil` to pick the correct OS-specific tmp dir?

hack/download-integration-test-binaries.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# Close stdin
+exec 0<&-
+
+ROOT_DIR="$(git rev-parse --show-toplevel)"
+
+ETCD_VERSION=v3.2.10
+ETCD_URL="https://storage.googleapis.com/etcd"
+
+KUBE_VERSION_URL="https://storage.googleapis.com/kubernetes-release/release/stable-1.10.txt"
```
munnerz, line 23: Can we pin this to a specific patch release of 1.10? Or otherwise, clearly print out the exact version returned here in the test logs, so we can debug failures caused by patch version changes.

test/integration/pilot_test.go
```diff
@@ -0,0 +1,200 @@
...
+	crdPath, err := filepath.Abs("testdata/pilot_crd.yaml")
+	require.NoError(t, err)
+	stdout, stderr, err := cli.Run(
+		"apply",
+		"--filename",
+		crdPath,
+	)
+	t.Log("stdout", stdout)
+	t.Log("stderr", stderr)
+	require.NoError(t, err)
```
munnerz, line 110: From here up to the start of the function should be in some shared function - this is code that is not specific to this test, or Cassandra.

Also, we should add some comments to this explaining *why* we are registering a CRD, and what is and isn't okay to test whilst using a CRD (because a lot of code for e.g. validation won't be run, we should clearly define bounds for what is and isn't okay to test). As we spoke about before, I think we should look to remove this code ASAP in favour of launching navigator-apiserver itself.

Also, given what we're trying to test here, would it be possible to implement this whole test as a unit test? I'm not sure why we need to bring up the entire process to verify that a config file is written ❓

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -45,20 +43,20 @@ func NewPilot(opts *PilotOptions) (*Pilot, error) {
...
+	//	"io.jetstack.cassandra.KubernetesSeedProvider", -1)
 
-	err = ioutil.WriteFile(cfgPath, []byte(newContents), 0)
-	if err != nil {
-		return nil, err
-	}
+	// err = ioutil.WriteFile(cfgPath, []byte(newContents), 0)
+	// if err != nil {
+	//	return nil, err
+	// }
```
munnerz: How come this is commented out? 🤔

pkg/config/config.go
```diff
@@ -0,0 +1,84 @@
...
+
+type config struct {
+	*viper.Viper
+}
+
+type Interface interface {
+	Get(key string) interface{}
+	Set(key string, value interface{})
+	Unset(key string)
+	WriteConfig() error
```
munnerz, line 32: Can this method be changed to accept an `io.Writer` as an argument?

The `New` function can also be changed to accept an `io.Read`, and we can then make the `newFrom` variants handle obtaining an `io.Reader` from a particular filepath.

pkg/config/config.go
```diff
@@ -0,0 +1,84 @@
...
+// A backup is made, if the file already exists.
+func (c *config) WriteConfig() error {
+	path := c.ConfigFileUsed()
+	backupPath := path + BackupSuffix
+	err := shutil.CopyFile(path, backupPath, false)
+	if err != nil && !os.IsNotExist(err) {
+		return errors.Wrap(err, "unable to create backup file")
+	}
+	return c.Viper.WriteConfig()
+}
```
munnerz, line 84: We can remove this backup file handling logic imo - it should be up to the caller to either overwrite or not overwrite a file as required.

pkg/pilot/cassandra/v3/pilot.go
```diff
@@ -127,3 +124,68 @@ func (p *Pilot) LivenessCheck() error {
...
+	cfg.Set(
+		"seed_provider",
+		[]map[string]interface{}{{
+			"class_name": CassSeedProvider,
+			"parameters": []interface{}{
+				map[string]interface{}{
+					"seeds": "",
+				},
+			},
+		}},
```
munnerz, line 158: 🤔 Is this a slice, of maps, of strings to interfaces? (of which one of those elements is a slice of interfaces, containing a map from strings to interfaces?!)

I guess this is just because of how Viper works (due to lack of generics), and how the YAML is structured, but... wow

docs/quick-start/cassandra-cluster.yaml
```diff
@@ -31,3 +31,5 @@ spec:
   pilotImage:
     repository: "quay.io/jetstack/navigator-pilot-cassandra"
     tag: "v0.1.0"
+  securityContext:
+    runAsUser: 999
```
munnerz, line 35: As per discussion earlier - what happens if this field is not specified on a Cassandra resource?

If we are not careful, this PR could become a breaking change as older versions of Navigator handled setting the uid automatically (i.e. via the entrypoint script) before forking to the cassandra subprocess.

docs/cassandra.rst
```diff
@@ -240,6 +242,37 @@ Navigator will add C* nodes, one at a time, until the desired number of nodes is
...
+
+By default, Navigator will use the `Cassandra Docker images from DockerHub <https://hub.docker.com/_/cassandra/>`_.
+It will use an image with a tag matching the supplied ``CassandraCluster.Spec.Version`` field.
+If you prefer to use your own container image you should configure the ``CassandraCluster.Spec.Image`` fields.
+
+Navigator installs a ``navigator-pilot-cassandra`` executable into each Pod at the path ``/pilot``.
+This ``pilot`` process connects to the API server to:
+get extra configuration settings,
+report the status of this C* node, and to
+perform leader election of a single pilot in the cluster.
```
munnerz, line 256: (nit) Bulleted list 😄

docs/cassandra.rst
```diff
@@ -240,6 +242,37 @@ Navigator will add C* nodes, one at a time, until the desired number of nodes is
...
+
+The ``pilot`` overrides the following keys in the default ``/etc/cassandra/cassandra.yaml`` file:
+
+* ``cluster_name``: This will be set to match the name of the corresponding ``CassandraCluster`` resource in the API server.
```
munnerz, line 260: Just a thought: is this safe to do, given two clusters may have the same name (in two different namespaces)

docs/cassandra.rst
```diff
@@ -240,6 +242,37 @@ Navigator will add C* nodes, one at a time, until the desired number of nodes is
...
+* ``listen_address`` / ``listen_interface`` / ``broadcast_address`` / ``rpc_address`` / ``broadcast_rpc_address``:  These keys will be set to ``null``.
+  This ensures that Cassandra process listens and communicates using the IP address currently associated with the fully qualified domain name of the Pod.
+  This is important if the Pod is moved to another node and is assigned a different IP address.
+  Removing these settings from the Configuration file ensures that Cassandra uses the most recent IP address that Kubernetes has assigned to the Pod and that other C* nodes in the cluster are notified of the change of IP address.
+* ``seed_provider``: This is set to ``io.jetstack.cassandra.KubernetesSeedProvider`` which allows Cassandra to look up the seed IP addresses from a Kubernetes service.
+  The ``seed_provider.*.seeds`` sub key will be emptied.
+  This is to avoid the risk of nodes mistakenly joining the cluster as seeds if the seed provider service is temporarily unavailable.
+
+The ``pilot`` also overwrites ``cassandra-rackdc.properties`` with values derived from the ``CassandraCluster.Spec.Nodepools`` (see :ref:`availability-zones-cassandra`).
```
munnerz, line 269: `Nodepools` -> `NodePools`

Gopkg.lock
```diff
@@ -301,18 +343,54 @@
...
+  name = "github.com/spf13/viper"
+  packages = ["."]
+  revision = "b5e8006cbee93ec955a89ab31e0e3ce3204f3736"
+  version = "v1.0.2"
+
+[[projects]]
+  name = "github.com/stretchr/testify"
+  packages = ["assert","require"]
+  revision = "12b6f73e6084dad08a7c6e575284b177ecafbc71"
+  version = "v1.2.1"
```
munnerz, line 386: How come we need testify? We've not used it in any of our other tests so far

Gopkg.lock
```diff
@@ -97,6 +97,12 @@
   packages = ["."]
   revision = "944e07253867aacae43c04b2e6a239005443f33a"
 
+[[projects]]
+  name = "github.com/fsnotify/fsnotify"
+  packages = ["."]
+  revision = "c2828203cd70a50dcccfb2761f8b1f8ceef9a8e9"
+  version = "v1.4.7"
```
munnerz, line 104: Any idea what introduces a need for fsnotify? 🤔 viper?

Issue #351, opened 2018-05-08
Cassandra E2E tests fail on a multi-node cluster because we use hostpath volumes


wallrj:
In the E2E tests, we configure Cassandra nodes with hostpath volumes.
These work on minikube because it only creates a single node.

On a multi-node cluster such as created by kubeadm-dind-cluster, Cassandra Pods may get  moved between nodes and in this case new (empty) hostpath volumes are created on each node and data appears to be lost.

Instead we could use persistent local storage PVs which will ensure that pods are only scheduled to nodes with the PV.

#330 has some example code for creating persistent local volumes for E2E tests.

/kind bug

wallrj:
Related to https://github.com/jetstack/navigator/issues/359

PR #366, opened 2018-06-05
Use local volumes instead of hostpath volumes

wallrj:
* Use `local` volumes rather than `hostpath` volumes in our e2e tests.
* I need this for #347, where I'm bypassing the Docker image entrypoint script.
* The entrypoint script would normally run `chown` to make `cassandra` user the owner of `/var/lib/cassandra`, and then exec `su cassandra cassandra`
* But Kubernetes can take care of that using `PodSecurityContext` and `SecurityContext` `RunAsUser` and `FSGroup`.
* Unfortunately, `FSGroup` doesn't work for HostPath volumes (https://github.com/kubernetes/kubernetes/pull/39438)
* `local` volume provisioner does apply `FSGroup` when mounting the volume.
* This change also allows e2e tests to pass on multi-node clusters, as described in #351 
* Although there's currently no way to enable PersistentLocalVolumes feature gate on kubeadm-dind clusters :-( (https://github.com/kubernetes-sigs/kubeadm-dind-cluster/issues/139)

Fixes: #351 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/libe2e.sh
+++ b/hack/libe2e.sh
@@ -253,6 +253,16 @@ function kube_create_pv() {
     local name="${1}"
     local capacity="${2}"
     local storage_class="${3}"
+    local schedulable_nodes
+    schedulable_nodes=$(
+        kubectl get nodes \
+                --output \
+                'jsonpath={range $.items[*]}{.metadata.name} {.spec.taints[*].effect}{"\n"}{end}' \
+            | grep -v NoSchedule)
+    local node_name
+    node_name=$(python -c 'import random,sys; print(random.choice(sys.argv[1:]))' $schedulable_nodes)
+
+    local path="hostpath_pvs/${name}/"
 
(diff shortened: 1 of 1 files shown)


hack/libe2e.sh
```diff
@@ -253,6 +253,16 @@ function kube_create_pv() {
...
     local capacity="${2}"
     local storage_class="${3}"
+    local schedulable_nodes
+    schedulable_nodes=$(
+        kubectl get nodes \
+                --output \
+                'jsonpath={range $.items[*]}{.metadata.name} {.spec.taints[*].effect}{"\n"}{end}' \
+            | grep -v NoSchedule)
+    local node_name
+    node_name=$(python -c 'import random,sys; print(random.choice(sys.argv[1:]))' $schedulable_nodes)
```
kragniz: can we assume python's going to be installed?

